PR #14, opened 2020-07-06
PMM-6129 Add service stub.

AlekSi:


service/cluster/xtra_db_cluster.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+// New returns new Service instance.
+func New() *Service {
+	return new(Service)
+}
+
+// ListXtraDBClusters returns a list of XtraDB clusters.
+func (s *Service) ListXtraDBClusters(ctx context.Context, req *controllerv1beta1.ListXtraDBClustersRequest) (*controllerv1beta1.ListXtraDBClustersResponse, error) {
+	return nil, status.Errorf(codes.Unimplemented, "This method is not implemented yet.")
```
AlekSi, line 39: @askomorokhov This error should reference i18n key instead of using hard-coded message

PR #16, opened 2020-07-08
PMM-6133 Add Basic i18n support.

askomorokhov:
https://jira.percona.com/browse/PMM-6133

Usage:
 - run `make gen` to collect text and marge with translation.
 - update catalog/locales/en/messages.gotext.json with appropriate translation.

Test:
 run in https://github.com/percona-platform/dbaas-api local clone:
```
$(DOCKER_RUN_CMD) prototool grpc --address host.docker.internal:20201  --method percona.platform.dbaas.controller.v1beta1.XtraDBClusterAPI/ListXtraDBClusters --data '{"kube_auth": {"kubeconfig": "122"}}'`
$(DOCKER_RUN_CMD) prototool grpc --address host.docker.internal:20201  --method percona.platform.dbaas.controller.v1beta1.XtraDBClusterAPI/CreateXtraDBCluster --data '{"kube_auth": {"kubeconfig": "122"}, "name": "cluster1"}'
```


it-percona:
[CLA assistant check](https://cla-assistant.percona.com:443/percona-platform/dbaas-controller?pullRequest=16) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.percona.com:443/percona-platform/dbaas-controller?pullRequest=16) before we can accept your contribution.<br/><hr/>**Andrii Skomorokhov** seems not to be a GitHub user. You need a GitHub account to be able to sign the CLA. If you have already a GitHub account, please [add the email address used for this commit to your account](https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user).<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.percona.com:443/check/percona-platform/dbaas-controller?pullRequest=16) it.</sub>

catalog/locales/en/out.gotext.json
```diff
@@ -0,0 +1,30 @@
...
+                }
+            ]
+        },
+        {
+            "id": "This method is not implemented yet.",
+            "message": "This method is not implemented yet.",
+            "translation": ""
+        }
+    ]
+}
```
AlekSi, line 30: Can we add \n to those two files?

catalog/README.md
```diff
@@ -0,0 +1,9 @@
+# Catalog
+
+This package holds i18n for DBaaS Controller.
+
+[gotext](https://pkg.go.dev/golang.org/x/text@v0.3.3/cmd/gotext?tab=doc)
+
+To merge translations and generate catalog use:
+
+    make gen
```
AlekSi, line 9: It is not clear what file should be updated manually with translations, why we need two files, what should be done after translation changed, etc.

catalog/locales/en/out.gotext.json
askomorokhov: I do not see positivities to do this in go text
https://github.com/golang/text/blob/23ae387dee1f90d29a23c0e87ee0b46038fbed0e/message/pipeline/pipeline.go#L362

should we format it using external tools like `jq` ?

AlekSi: Well, I would use `cat`, `echo` or something :)

catalog/README.md
```diff
@@ -0,0 +1,19 @@
+# Catalog
+
+This package holds i18n for DBaaS Controller.
+
+[gotext](https://pkg.go.dev/golang.org/x/text@v0.3.3/cmd/gotext?tab=doc)
+
+To translate, we need to follow the next steps.
+
+- Run `make gen` to collect all messages from source code; it also merges them with existing translation in `messages.gotext.json` and writes them into `out.gotext.json`.
```
AlekSi, line 9: Shouldn't we call `gotext extract` for that?

catalog/README.md
```diff
@@ -0,0 +1,19 @@
...
+- As a result, we will receive all new (not translated messages) and all old (translated messages from `messages.gotext.json`) in `out.gotext.json`.
+- In the next step, we need to copy `out.gotext.json` into messages.gotext.json and fill in all required translation values.
```
AlekSi, line 11: Why we need to copy it manually? Can we automate this?
Do we also need to keep both files?

catalog/README.md
askomorokhov: For this particular step `gotext extract` can be as well.  BTW `gotext update` also extract all messages plus generate code. So we can use `gotext update` for both steps.

catalog/README.md
askomorokhov: Good point.
I don't have a "right" answer for this question, I did not find any best practice for this tool as it is relatively new, and our case is particular - translation from "English" to English.
When I experimented with the translation, I found it useful to have both these files: in case `messages.gotext.json` was mess up, I always could refer to `out.gotext.json` as a source of truth. 

A possible case: `messages.gotext.json` should not be congruent to `out.gotext.json`, so I guess it will be handy  to copy only needed part of `out.gotext.json`

AlekSi: We have git to fix a mess if we make it.

In what case only a copy of the file should be copied?

askomorokhov: > We have git to fix a mess if we make it.

not a git in this case but `make gen` - that take much more time then take a look at out.gotext.json in the next window split.

> In what case only a copy of the file should be copied?

For example when we do translation ahead
```
--- a/catalog/locales/en/messages.gotext.json
+++ b/catalog/locales/en/messages.gotext.json
@@ -25,6 +25,11 @@
             "id": "This method is not implemented yet.",
             "message": "This method is not implemented yet.",
             "translation": ""
+        },
+        {
+            "id": "TBD#1",
+            "message": "TBD#1",
+            "translation": "Some translation"
         }
     ]
 }

```

AlekSi: >> We have git to fix a mess if we make it.
> not a git in this case but make gen 

I meant `git reset --hard` / `git checkout`.

Ahead of what? The actual code that uses that text? It's hard to image that situation in practice, TBH.

Let's keep one file in git – that way it is much harder to mess things up while holding them wrong.

askomorokhov: With `git reset --hard / git checkout` - you lose parsed/generated messages.
and if code base is big enough-  re-run `make gen` may take ~1-5 min - here is my point - but, ok, it can be worked around with other tools.

So, your suggestion is:
 - add to target gen `mv catalog/locales/en/{out,messages}.gotext.json`
 - remove `echo >> catalog/locales/en/out.gotext.json`
 - add `catalog/locales/en/out.gotext.json` to `.gitignore`

Correct?

AlekSi: > you lose parsed/generated messages

No, if they are committed in git. That's the whole point of using it.

IIUC – yes. We can always make it more complex when/if we need it. Making things simpler is typically harder :)

.gitignore
```diff
@@ -4,3 +4,4 @@
 /bin/
 /debugdata/
 *.out
+/catalog/locales/*/out.gotext.json
```
AlekSi, line 7: Why we ignore it?

catalog/README.md
```diff
@@ -0,0 +1,21 @@
+# Catalog
+
+This package holds i18n for dbaas-controller.
+
+[gotext](https://pkg.go.dev/golang.org/x/text@v0.3.3/cmd/gotext?tab=doc)
+
+To translate dbaas-controller's messages, we need to follow the next steps.
+
+- Use shared `golang.org/x/text/message.Printer` object's `Sprintf` method for messages that should be translated.
+  Do not use `Sprint` as it will not be collected on the next step.
```
AlekSi, line 10: @askomorokhov did you encounter this?

askomorokhov: Going to remove - missed this old item.

AlekSi: I just added it. But can you verify that problem?

askomorokhov: Oh, it was removed. Good. Yes, I met it  - but as I figure out later it was caused not by Sprint but by another error.

PR #20, opened 2020-07-18
PMM-6220 Use Docker driver

AlekSi:


gen1us2k:
We use EKS for our pipelines now.

PR #22, opened 2020-07-22
PMM-5895 Basic kubernetes client.

BupycHuk:


service/k8sclient/k8client.go
```diff
@@ -0,0 +1,280 @@
...
+
+		cluster := Cluster{
+			Status: "deleting",
+			Kind:   kind,
+			Name:   clusterName,
+		}
+		res = append(res, cluster)
+
+		exists[clusterName] = struct{}{}
+	}
```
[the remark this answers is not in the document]
BupycHuk: I ran gofumpt locally, but it's still here.

.golangci.yml
```diff
@@ -27,6 +27,7 @@ linters:
     - gomnd     # too annoying
     - lll       # gRPC types are long
     - wsl       # too annoying
+    - unused    # too many false positives
```
AlekSi, line 30: It actually doesn't have false positives: we don't use that code outside of the tests, and linter is doing a good job checking that we don't keep unused code.

Let's keep it there, but I will revert that change after merging.

service/k8sclient/k8client.go
```diff
@@ -0,0 +1,273 @@
...
+	Name string
+	Size int32
+}
+
+// Cluster contains information related to cluster.
+type Cluster struct {
+	Name   string
+	Kind   ClusterKind
+	Size   int32
+	Status string // FIXME
```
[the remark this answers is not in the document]
AlekSi: This should be an own type too

service/k8sclient/k8client.go
```diff
@@ -0,0 +1,273 @@
...
+	}
+}
+
+// Cleanup removes temporary files created by that object.
+func (c *K8Client) Cleanup() {
+	c.kubeCtl.Cleanup()
+}
+
+// ListClusters returns list of clusters and their statuses.
+func (c *K8Client) ListClusters(ctx context.Context) ([]Cluster, error) {
```
AlekSi, line 79: There should two method for two types of clusters – they have different properties like ProxySQL settings for PXC

service/k8sclient/k8client.go
```diff
@@ -0,0 +1,273 @@
...
+	res := append(perconaXtraDBClusters, deletingClusters...)
+
+	return res, nil
+}
+
+// CreateXtraDBCluster creates Percona XtraDB cluster with provided parameters.
+func (c *K8Client) CreateXtraDBCluster(ctx context.Context, params *XtraDBParams) error {
+	res := &pxc.PerconaXtraDBCluster{
+		TypeMeta: meta.TypeMeta{
+			APIVersion: "pxc.percona.com/v1-4-0",
```
AlekSi, line 100: That should be a named constant

service/k8sclient/kubectl/kubectl.go
```diff
@@ -0,0 +1,130 @@
...
+// It returns `errNotFound` if object is not found.
+func (k *KubeCtl) Get(ctx context.Context, kind string, name string, res interface{}) error {
+	args := []string{"get", "-o=json", kind}
+	if name != "" {
+		args = append(args, name)
+	}
+
+	stdout, stderr, err := k.run(ctx, args, nil)
+	if err != nil {
+		if strings.Contains(string(stderr), "not found") { // FIXME
```
AlekSi, line 80: Can we do better?

service/k8sclient/kubectl/kubectl.go
```diff
@@ -0,0 +1,130 @@
...
+	var outBuf bytes.Buffer
+	var errBuf bytes.Buffer
+	cmd := exec.CommandContext(ctx, args[0], args[1:]...) //nolint:gosec
+	pdeathsig.Set(cmd, unix.SIGKILL)
+	cmd.Stdin = &inBuf
+	cmd.Stdout = &outBuf
+	cmd.Stderr = &errBuf
+	err := cmd.Run()
+
+	k.l.Debugf(outBuf.String()) // FIXME
```
AlekSi, line 127: That will not work if output contains `%s` or something similar. We can't use XXXf functions as just XXX.

service/k8sclient/kubectl/kubectl.go
BupycHuk: Removed it since we don't need it anymore.

service/k8sclient/kubectl/kubectl.go
BupycHuk: Added new Debug method.

AlekSi: Why add more methods to interface and implementations instead of using `Debugf("%s", outBuf)`?

BupycHuk: ok, anything else?

BupycHuk: Added other methods, since we already have the same usages for Fatal, Info and Debug.

PR #34, opened 2020-08-06
PMM-6348 Handle kubectl versions

askomorokhov:
https://jira.percona.com/browse/PMM-6348

FB: https://github.com/Percona-Lab/pmm-submodules/pull/1002

### DIFF
--- a/service/k8sclient/k8client.go
+++ b/service/k8sclient/k8client.go
@@ -95,9 +95,12 @@ type K8Client struct {
 // NewK8Client returns new K8Client object.
 func NewK8Client(ctx context.Context) (*K8Client, error) {
 	kubeCtl, err := kubectl.NewKubeCtl(ctx)
+	if err != nil {
+		return nil, err
+	}
 	return &K8Client{
 		kubeCtl: kubeCtl,
-	}, err
+	}, nil
 }
 
 // Cleanup removes temporary files created by that object.
--- a/service/k8sclient/k8client.go
+++ b/service/k8sclient/k8client.go
@@ -30,7 +30,6 @@ import (
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/percona-platform/dbaas-controller/service/k8sclient/kubectl"
-	"github.com/percona-platform/dbaas-controller/utils/logger"
 )
 
 // ClusterKind is a kind of a cluster.
@@ -94,10 +93,11 @@ type K8Client struct {
 }
 
 // NewK8Client returns new K8Client object.
-func NewK8Client(logger logger.Logger) *K8Client {
+func NewK8Client(ctx context.Context) (*K8Client, error) {
+	kubeCtl, err := kubectl.NewKubeCtl(ctx)
 	return &K8Client{
-		kubeCtl: kubectl.NewKubeCtl(logger),
-	}
+		kubeCtl: kubeCtl,
+	}, err
 }
 
 // Cleanup removes temporary files created by that object.
--- a/service/k8sclient/kubectl/kubectl.go
+++ b/service/k8sclient/kubectl/kubectl.go
@@ -53,7 +53,6 @@ func NewKubeCtl(ctx context.Context) (*KubeCtl, error) {
 	// Handle kubectl versions
 	cmd, err := getKubectlCmd(ctx)
 	if err != nil {
-		l.Debugf("Cannot find kubectl binary: %s", err)
 		return nil, err
 	}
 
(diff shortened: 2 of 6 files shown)


service/k8sclient/kubectl/kubectl.go
```diff
@@ -40,24 +48,106 @@ type KubeCtl struct {
...
-	cmd := []string{"dbaas-kubectl-1.16"}
-	path, err := exec.LookPath(cmd[0])
-	l.Debugf("path: %s, err: %v", path, err)
-	if e, ok := err.(*exec.Error); err != nil && ok && e.Err == exec.ErrNotFound {
-		cmd = []string{"minikube", "kubectl", "--"}
+	// Handle kubectl versions
+	cmd, err := getKubectlCmd()
+	if err != nil {
+		l.Debugf("Cannot find kubectl binary: %s", err)
+		return nil
```
AlekSi, line 55: Let's return that error and handle it in the caller

service/k8sclient/kubectl/kubectl.go
```diff
@@ -32,6 +35,11 @@ import (
 	"github.com/percona-platform/dbaas-controller/utils/logger"
 )
 
+const (
+	defaultKubectl = "dbaas-kubectl-1.16"
```
AlekSi, line 39: That's good, but let's temporary set it to `"minikube kubectl --"` with a FIXME comment

service/k8sclient/kubectl/kubectl.go
```diff
@@ -40,24 +48,106 @@ type KubeCtl struct {
...
+	kubectlCmdNames, err := selectCorrectKubectlVersions(versionsJSON)
+	if err != nil {
+		return nil, err
+	}
+
+	return lookupCorrectKubectlCmd(kubectlCmdNames)
+}
+
+// getVersions gets kubectl and Kubernetes cluster version.
+func getVersions(kubectlCmd []string) ([]byte, error) {
```
AlekSi, line 89: I would expect those functions to be defined in the same order as they used: lookupCorrectKubectlCmd, getVersions, selectCorrectKubectlVersions

service/k8sclient/kubectl/kubectl.go
```diff
@@ -40,24 +48,106 @@ type KubeCtl struct {
...
+	ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
+	defer cancel()
+
+	getVersionCmd := append(kubectlCmd, "version", "-o", "json")
+	cmd := exec.CommandContext(ctx, getVersionCmd[0], getVersionCmd[1:]...) //nolint:gosec
```
AlekSi, line 94: I think we should refactor `run` method to make it a function, and the use it everywhere instead of exec.Command/CommandContext

askomorokhov: Agree. Also thought about this.

service/k8sclient/kubectl/kubectl.go
askomorokhov: Reordered.

service/k8sclient/k8client.go
```diff
@@ -94,10 +93,11 @@ type K8Client struct {
...
 
 // NewK8Client returns new K8Client object.
-func NewK8Client(logger logger.Logger) *K8Client {
+func NewK8Client(ctx context.Context) (*K8Client, error) {
+	kubeCtl, err := kubectl.NewKubeCtl(ctx)
 	return &K8Client{
-		kubeCtl: kubectl.NewKubeCtl(logger),
-	}
+		kubeCtl: kubeCtl,
+	}, err
```
AlekSi, line 100: I think we should use slightly more traditional error handling there.
image

service/k8sclient/kubectl/kubectl.go
```diff
@@ -32,30 +34,116 @@ import (
...
-
-	cmd := []string{"dbaas-kubectl-1.16"}
-	path, err := exec.LookPath(cmd[0])
-	l.Debugf("path: %s, err: %v", path, err)
-	if e, ok := err.(*exec.Error); err != nil && ok && e.Err == exec.ErrNotFound {
-		cmd = []string{"minikube", "kubectl", "--"}
+	// Handle kubectl versions
+	cmd, err := getKubectlCmd(ctx)
+	if err != nil {
+		l.Debugf("Cannot find kubectl binary: %s", err)
```
AlekSi, line 56: In general, we should either handler errors (by logging them, for example), or return them, but not both. That error will be logged by an interceptor, so let's just return it.

PR #58, opened 2020-09-21
PMM-6457 Add kubectl of pmm-server.

askomorokhov:
https://jira.percona.com/browse/PMM-6457

ref:
- https://github.com/Percona-Lab/pmm-api-tests/pull/125
- https://github.com/percona-platform/dbaas-controller/pull/58
- https://github.com/percona/pmm-managed/pull/468
- https://github.com/percona/pmm/pull/582
- https://github.com/Percona-Lab/pmm-submodules/pull/1124

### DIFF
--- a/service/k8sclient/kubectl/kubectl.go
+++ b/service/k8sclient/kubectl/kubectl.go
@@ -37,7 +37,6 @@ import (
 )
 
 const (
-	kubeconfigFileName      = "kubeconfig.json"
 	defaultPmmServerKubectl = "/opt/dbaas-tools/bin/kubectl-1.16"
 	defaultDevEnvKubectl    = "minikube kubectl --"
 )
--- a/service/k8sclient/kubectl/kubectl.go
+++ b/service/k8sclient/kubectl/kubectl.go
@@ -25,7 +25,6 @@ import (
 	"io/ioutil"
 	"os"
 	"os/exec"
-	"path"
 	"strconv"
 	"strings"
 
@@ -45,9 +44,9 @@ const (
 
 // KubeCtl wraps kubectl CLI with version selection and kubeconfig handling.
 type KubeCtl struct {
-	l      logger.Logger
-	cmd    []string
-	tmpDir string
+	l              logger.Logger
+	cmd            []string
+	kubeconfigPath string
 }
 
 // NewKubeCtl creates a new KubeCtl object with a given logger.
@@ -56,7 +55,8 @@ func NewKubeCtl(ctx context.Context, kubeconfig string) (*KubeCtl, error) {
 	l = l.WithField("component", "kubectl")
 
 	// Firstly lookup default kubectl to get Kubernetes Server version.
-	defaultKubectl, err := getDefaultKubectlCmd()
+	defKubectls := []string{defaultPmmServerKubectl, defaultDevEnvKubectl}
+	defaultKubectl, err := lookupCorrectKubectlCmd(nil, defKubectls)
 	if err != nil {
 		return nil, err
 	}
@@ -70,7 +70,7 @@ func NewKubeCtl(ctx context.Context, kubeconfig string) (*KubeCtl, error) {
 	}
 
 	// Handle kubeconfig.
-	tmpDir, kubeconfigPath, err := saveKubeconfig(kubeconfig)
+	kubeconfigPath, err := saveKubeconfig(kubeconfig)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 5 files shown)


service/k8sclient/kubectl/kubectl.go
```diff
@@ -38,8 +38,9 @@ import (
 )
 
 const (
-	// FIXME: change to a real default kubectl.
-	defaultKubectl = "minikube kubectl --"
+	kubeconfigFileName      = "kubeconfig.json"
+	defaultPmmServerKubectl = "/opt/dbaas-tools/bin/kubectl-1.16" // not in $PATH yet
```
AlekSi, line 42: It will not be in PATH. If say otherwise somewhere, please update that place.

service/k8sclient/kubectl/kubectl.go
```diff
@@ -38,8 +38,9 @@ import (
 )
 
 const (
-	// FIXME: change to a real default kubectl.
-	defaultKubectl = "minikube kubectl --"
+	kubeconfigFileName      = "kubeconfig.json"
```
AlekSi, line 41: Is it actually used somewhere?

service/k8sclient/kubectl/kubectl.go
```diff
@@ -213,6 +233,7 @@ func (k *KubeCtl) Apply(ctx context.Context, res meta.Object) error {
 // Delete executes `kubectl delete` with given resource.
 func (k *KubeCtl) Delete(ctx context.Context, res meta.Object) error {
 	_, err := run(ctx, k.cmd, []string{"delete", "-f", "-"}, res)
+
```
AlekSi, line 236: Please disable linter that asks you do that

service/k8sclient/kubectl/kubectl.go
askomorokhov: Looks like I incorrectly understood your words. I'll remove this comment.

service/k8sclient/kubectl/kubectl.go
askomorokhov: Only here https://github.com/percona-platform/dbaas-controller/blob/main/service/k8sclient/kubectl/kubectl.go#L97 and in test https://github.com/percona-platform/dbaas-controller/blob/main/service/k8sclient/kubectl/kubectl_test.go#L54-L61

Do we need to change it to simply `ioutil.TempFile`?

service/k8sclient/kubectl/kubectl.go
AlekSi: No, it was supposed to be in $PATH, but that was changed

service/k8sclient/kubectl/kubectl.go
```diff
@@ -122,10 +140,10 @@ func getKubectlCmd(ctx context.Context) ([]string, error) {
 		return nil, err
 	}
 
-	return lookupCorrectKubectlCmd(kubectlCmdNames)
+	return lookupCorrectKubectlCmd(defaultKubectl, kubectlCmdNames)
 }
 
-func lookupCorrectKubectlCmd(kubectlCmdNames []string) ([]string, error) {
+func lookupCorrectKubectlCmd(defaultKubectl, kubectlCmdNames []string) ([]string, error) {
```
BupycHuk, line 146: Why can't we just add `defaultKubectl` to `kubectlCmdNames`?

service/k8sclient/kubectl/kubectl.go
```diff
@@ -134,15 +152,16 @@ func lookupCorrectKubectlCmd(kubectlCmdNames []string) ([]string, error) {
...
 	// if none found (pass empty kubectlCmdNames) use default version of kubectl.
-	return strings.Split(defaultKubectl, " "), nil
+	return defaultKubectl, nil
 }
 
 // getVersions gets kubectl and Kubernetes cluster version.
-func getVersions(ctx context.Context, kubectlCmd []string) ([]byte, error) {
-	versionsJSON, err := run(ctx, kubectlCmd, []string{"version", "-o", "json"}, nil)
+func getVersions(ctx context.Context, kubectlCmd []string, kubeconfigPath string) ([]byte, error) {
+	versionsJSON, err := run(ctx, kubectlCmd, []string{"version", fmt.Sprintf("--kubeconfig=%s", kubeconfigPath), "-o", "json"}, nil)
```
BupycHuk, line 160: :+1: I also fixed it on my branch.

BupycHuk: Isn't it weird that we can pass flag with empty value?

service/k8sclient/kubectl/kubectl.go
```diff
@@ -104,15 +110,27 @@ func saveKubeconfig(kubeconfig string) (string, string, error) {
...
 }
 
-// getKubectlCmd gets correct version of kubectl binary for Kubernetes cluster.
-func getKubectlCmd(ctx context.Context) ([]string, error) {
-	// Firstly lookup default kubectl to get Kubernetes Server version.
-	kubectlCmd, err := lookupCorrectKubectlCmd([]string{defaultKubectl})
-	if err != nil {
-		return nil, err
+func getDefaultKubectlCmd() ([]string, error) {
+	kubectlPath, errPmmServerKubectl := exec.LookPath(defaultPmmServerKubectl)
```
BupycHuk, line 114: Why can't we just use `lookupCorrectKubectlCmd` with two values in slice instead of this method?

BupycHuk (changes requested):

service/k8sclient/kubectl/kubectl.go
askomorokhov: it should not pass so far if "kubeconfig" is empty:
https://github.com/percona-platform/dbaas-controller/blob/c8ff486b9a477a9999c5d5c2c52ebeedf9c24944/service/k8sclient/kubectl/kubectl.go#L64-L70

service/k8sclient/kubectl/kubectl.go
askomorokhov: Yeah, I also thought in such a way and even started to implement it.
But then I refactored it to separate them, follows a rule: one task one function.
We have two steps:
1. find default kubectl;
2. find the latest valid kubectl version for server
and this makes the logic more evident if we have a dedicated function to each step.


In opposite: joining it into one function introduce an additional hidden dependency on order on the lookup list with the default and introduce additional complexity for local dev env (as it is not a real path but "alias").

artemgavrilov (approved):

service/k8sclient/kubectl/kubectl.go
BupycHuk: ok

service/k8sclient/kubectl/kubectl.go
BupycHuk: I didn't suggest to replace it to `getKubectlCmd`, I suggested to replace it to `lookupCorrectKubectlCmd`

service/k8sclient/kubectl/kubectl.go
BupycHuk: In this case if we can't find `kubectl` at all we can return error.

service/k8sclient/kubectl/kubectl.go
AlekSi: Yeah. It doesn't even have to be JSON file – YAML will work, and that's what most people will use

BupycHuk (changes requested):
There are some questions which weren't answered yet.

service/k8sclient/kubectl/kubectl.go
askomorokhov: Fixed.

service/k8sclient/kubectl/kubectl.go
askomorokhov: Fixed.

AlekSi:
@BupycHuk @askomorokhov https://github.com/percona-platform/dbaas-controller/pull/60#discussion_r499559719

service/k8sclient/kubectl/kubectl.go
```diff
@@ -122,27 +123,33 @@ func getKubectlCmd(ctx context.Context) ([]string, error) {
...
 	}
 
-	// if none found (pass empty kubectlCmdNames) use default version of kubectl.
-	return strings.Split(defaultKubectl, " "), nil
+	if defaultKubectl == nil {
+		return nil, errors.Errorf("cannot find kubectl: %v, %v", defaultKubectl, kubectlCmdNames)
+	}
+
+	// if none found and default is not empty use default version of kubectl.
+	return defaultKubectl, nil
```
BupycHuk, line 143: What's is the reason to use `defaultKubectl`? if we got kubernetes cluster with version 1.19, but we have only kubectl-1.16 can we be sure that everything will work correctly? And should we try?

BupycHuk: @AlekSi what do you think?

AlekSi: We should have _some_ fallback if we could not pick better version

BupycHuk: ok

BupycHuk (approved):

AlekSi (approved):

PR #60, opened 2020-10-01
PMM-6433 check connection to kubernetes cluster.

BupycHuk:
https://github.com/percona-platform/dbaas-api/pull/28
https://github.com/Percona-Lab/pmm-submodules/pull/1135
https://github.com/percona/pmm-managed/pull/474
https://github.com/Percona-Lab/pmm-api-tests/pull/128


service/k8sclient/kubectl/kubectl.go
```diff
@@ -56,31 +57,38 @@ func NewKubeCtl(ctx context.Context, kubeconfig string) (*KubeCtl, error) {
...
 			cmd: cmd,
 		}, nil
 	}
 
+	var tmpDir, kubeconfigPath string
+	var err error
+
 	// Handle kubeconfig.
-	tmpDir, kubeconfigPath, err := saveKubeconfig(kubeconfig)
+	tmpDir, kubeconfigPath, err = saveKubeconfig(kubeconfig)
```
artemgavrilov, line 76: Why did you replace `:=` with explicit variable declarations?

service/k8sclient/kubectl/kubectl.go
```diff
@@ -105,41 +113,48 @@ func saveKubeconfig(kubeconfig string) (string, string, error) {
...
 	if err != nil {
 		return nil, err
 	}
 
-	versionsJSON, err := getVersions(ctx, kubectlCmd)
+	versionsJSON, err := getVersions(ctx, kubectlCmd, kubeConfigPath)
 	if err != nil {
 		return nil, err
 	}
+	fmt.Println(string(versionsJSON))
```
artemgavrilov, line 127: Debug?

service/k8sclient/kubectl/kubectl.go
```diff
@@ -105,41 +113,48 @@ func saveKubeconfig(kubeconfig string) (string, string, error) {
...
+		cmd := strings.Split(kubectlCmdName, " ")
+		kubectlPath, err := exec.LookPath(cmd[0])
 		if err == nil {
-			return []string{kubectlPath}, nil
+			cmd[0] = kubectlPath
+			return cmd, nil
 		}
 	}
 
 	// if none found (pass empty kubectlCmdNames) use default version of kubectl.
```
artemgavrilov, line 147: That comment is not valid anymore.

service/k8sclient/kubectl/kubectl.go
BupycHuk: oops, I was moving lines up and down and then forgot to remove :)

service/k8sclient/kubectl/kubectl.go
BupycHuk: yeah, will remove

artemgavrilov (approved):

service/cluster/kubernetes_cluster.go
```diff
@@ -0,0 +1,48 @@
...
+// NewKubernetesClusterService returns new KubernetesClusterService instance.
+func NewKubernetesClusterService(p *message.Printer) *KubernetesClusterService {
+	return &KubernetesClusterService{p: p}
+}
+
+// CheckKubernetesClusterConnection checks connection with kubernetes cluster.
+func (k KubernetesClusterService) CheckKubernetesClusterConnection(ctx context.Context, req *controllerv1beta1.CheckKubernetesClusterConnectionRequest) (*controllerv1beta1.CheckKubernetesClusterConnectionResponse, error) {
+	k8Client, err := k8sclient.NewK8Client(ctx, req.KubeAuth.Kubeconfig)
+	if err != nil {
+		return nil, status.Error(codes.Internal, k.p.Sprintf("Unable to connect to kubernetes cluster: %s", err))
```
AlekSi, line 44: FailedPrecondition seems to be a better fit. That's also what we use in pmm-managed connection checkers

service/cluster/kubernetes_cluster_test.go
```diff
@@ -0,0 +1,64 @@
...
+	controllerv1beta1 "github.com/percona-platform/dbaas-api/gen/controller"
+	"github.com/sirupsen/logrus"
+	"github.com/stretchr/testify/require"
+	"golang.org/x/text/language"
+	"golang.org/x/text/message"
+)
+
+func TestKubernetesClusterServiceCheckConnection(t *testing.T) {
+	t.Run("Wrong kube config", func(t *testing.T) {
+		logrus.SetLevel(logrus.DebugLevel)
```
AlekSi, line 32: Do we need it?

service/cluster/kubernetes_cluster_test.go
```diff
@@ -0,0 +1,64 @@
...
+				{
+					"name": "minikube",
+					"context": {
+						"cluster": "minikube",
+						"user": "minikube"
+					}
+				}
+			],
+			"current-context": "minikube"
+		}`
```
AlekSi, line 57: Why we use JSON instead of YAML?

service/cluster/kubernetes_cluster_test.go
```diff
@@ -0,0 +1,64 @@
...
+		_, err := k.CheckKubernetesClusterConnection(context.TODO(), &controllerv1beta1.CheckKubernetesClusterConnectionRequest{
+			KubeAuth: &controllerv1beta1.KubeAuth{Kubeconfig: kubeConfig},
+		})
+		require.Error(t, err)
+		require.Regexp(t, "rpc error: code = Internal desc = Unable to connect to kubernetes cluster: .*", err.Error())
```
AlekSi, line 62: Let's port our gRPC error helper

service/cluster/xtra_db_cluster.go
```diff
@@ -37,18 +37,18 @@ var pxcStatesMap = map[k8sclient.ClusterState]controllerv1beta1.XtraDBClusterSta
 	k8sclient.ClusterStateDeleting: controllerv1beta1.XtraDBClusterState_XTRA_DB_CLUSTER_STATE_DELETING,
 }
 
-// Service implements methods of gRPC server and other business logic.
-type Service struct {
+// XtraDBClusterService implements methods of gRPC server and other business logic related to xtradb clusters.
```
AlekSi, line 40: XtraDB

service/cluster/kubernetes_cluster.go
AlekSi, line 44: Also, "Kubernetes", with a capital K

service/k8sclient/kubectl/kubectl.go
```diff
@@ -38,8 +38,9 @@ import (
 )
 
 const (
-	// FIXME: change to a real default kubectl.
-	defaultKubectl = "minikube kubectl --"
+	dbaasToolPath      = "/opt/dbaas-tools/bin"
+	defaultKubectl     = dbaasToolPath + "/kubectl-1.16"
+	minikubeDevKubectl = "minikube kubectl --"
```
AlekSi, line 43: Isn't it https://github.com/percona-platform/dbaas-controller/pull/58 ?

service/k8sclient/kubectl/kubectl.go
```diff
@@ -122,24 +127,30 @@ func getKubectlCmd(ctx context.Context) ([]string, error) {
...
 
-	return lookupCorrectKubectlCmd(kubectlCmdNames)
+	return lookupCorrectKubectlCmd(append(kubectlCmdNames, defaultKubectl, minikubeDevKubectl))
 }
 
 func lookupCorrectKubectlCmd(kubectlCmdNames []string) ([]string, error) {
 	for _, kubectlCmdName := range kubectlCmdNames {
-		kubectlPath, err := exec.LookPath(kubectlCmdName)
+		cmd := strings.Split(kubectlCmdName, " ")
+		kubectlPath, err := exec.LookPath(cmd[0])
```
AlekSi, line 136: LookPath will not work because `/opt/dbaas-tools/bin` is not in path, and should not be.

We also should add `/opt/dbaas-tools/bin` to cmd'e environment in `run` to handle iam-aws-authenticator and other tools

AlekSi (changes requested):
Shouldn't we also add API test?

service/k8sclient/kubectl/kubectl.go
BupycHuk: yes, it is, and I'll merge conflicts. I wasn't sure which one will be merged first.
This changes was required for my task as well.

service/k8sclient/kubectl/kubectl.go
BupycHuk: > If file contains a slash, it is tried directly and the PATH is not consulted. The result may be an absolute path or a path relative to the current directory.

I pass absolute path to the binary.

service/cluster/kubernetes_cluster_test.go
BupycHuk: Ok, but main issue here that we include path to temp file in the error message.

service/cluster/kubernetes_cluster_test.go
BupycHuk: Why not? We were using `-o json`, that's why I used the same format.

service/cluster/kubernetes_cluster_test.go
BupycHuk: Actually not, will remove.

service/cluster/kubernetes_cluster.go
BupycHuk: Ok

service/cluster/kubernetes_cluster_test.go
AlekSi: We use JSON for K8s API requests and responses, but pretty much everyone uses YAML for kubeconfig file. We can reasonably expect people to provide kubeconfig in YAML format. _If_ both JSON and YAML work now, then we can keep, but otherwise we should not force users to convert YAML to JSON (and using YAML will cover JSON too since YAML 1.2 is JSON superset).

service/cluster/kubernetes_cluster_test.go
AlekSi: Then let's port our helper with regex :) I think we have it somewhere

service/k8sclient/kubectl/kubectl.go
AlekSi: Then maybe we can just close the other one?

service/k8sclient/kubectl/kubectl.go
AlekSi: Yes, but we can't expect kubeconfig file to use `/opt/dbaas-tools/bin/iam-aws-authenticator` instead of just `iam-aws-authenticator`. So we should run `/opt/dbaas-tools/bin/kubectl-XXX` but with modified environment so `iam-aws-authenticator` can be found

BupycHuk: My answer was for the first, I'm totally agree with a second part related to iam-aws-authenticator.

AlekSi: So we use LookPath only to check that it exist?

AlekSi: Ah, and to lookup `minikube`. Got it. Maybe a comment would help :)

service/cluster/kubernetes_cluster_test.go
BupycHuk: I'll check if YAML works and if it is, will create separate test for it as well

askomorokhov: Just a note.
YAML should work as well. The issue could be to pass it from frontend to pmm-managed; YAML inside JSON, escaping not enough. Maybe a form- or base64- encoding/deconning is required

service/k8sclient/kubectl/kubectl.go
BupycHuk: No, I pulled changes from his PR and going to merge this PR after @askomorokhov PR.

service/cluster/kubernetes_cluster_test.go
BupycHuk: Checked it and it works :)

AlekSi: So `kubectl` always uses YAML parser for `kubeconfig` file, and since JSON is a subset of YAML 1.2, it just works

AlekSi: But I think we should use YAML both internally (in all tests, etc) and externally (in documentation, etc) simply because it is more common and tests more

service/cluster/kubernetes_cluster_test.go
```diff
@@ -0,0 +1,78 @@
...
+			KubeAuth: &controllerv1beta1.KubeAuth{Kubeconfig: kubeConfig},
+		})
+		require.Error(t, err)
+		AssertGRPCErrorRE(t, codes.FailedPrecondition, "Unable to connect to Kubernetes cluster: exit status 1", err)
+	})
+}
+
+// AssertGRPCErrorRE checks that actual error has expected gRPC error code, and error messages
+// matches expected regular expression.
+func AssertGRPCErrorRE(tb testing.TB, expectedCode codes.Code, expectedMessageRE string, actual error) {
```
AlekSi, line 69: Let's move it to some testutil package, not keep _right there_ and _exported_.

At very least, please un-export it

service/k8sclient/kubectl/kubectl.go
```diff
@@ -258,6 +259,10 @@ func run(ctx context.Context, kubectlCmd []string, args []string, stdin interfac
 	cmd.Stdin = &inBuf
 	cmd.Stdout = &outBuf
 	cmd.Stderr = &errBuf
+	cmd.Env = []string{
```
AlekSi, line 262: ⚠️ Sorry, that's incorrect. That completely replaces the environment, and we should add to it: get `os.Environ()`, iterate over it, find and update PATH, add to `cmd.Env`

AlekSi (changes requested):

tests/helpers.go
```diff
@@ -0,0 +1,38 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// AssertGRPCErrorRE checks that actual error has expected gRPC error code, and error messages
+// matches expected regular expression.
+func AssertGRPCErrorRE(tb testing.TB, expectedCode codes.Code, expectedMessageRE string, actual error) {
```
AlekSi, line 29: So we can use testutil's version there now?

AlekSi (approved):

tests/helpers.go
BupycHuk: I though one of restricts of having api tests in the same repository is not use packages outside of tests package, isn't it?

AlekSi: Yeah, good point. We probably should extract them into own repository to make it impossible :)

askomorokhov (approved):

PR #77, opened 2020-10-29
Check that API tests can be built

AlekSi:


AlekSi:
And do we have a ticket for actually running API tests somewhere? If not, let's create one and put it somewhere near the top of the backlog. Otherwise, they will be neglected.

BupycHuk (approved):

PR #78, opened 2020-10-30
PMM-6805 Exclude test package from check

percona-csalguero:
In go 1.15, AST analyzer in go-consistent reports that the test package is not being used in the main dbass module. It is true but it is a testing package so we can exclude it from checks.


AlekSi (changes requested):
Please add another line that actually checks tests module

AlekSi:
@BupycHuk That might be an argument to extract API tests into own repo

BupycHuk (approved):

AlekSi (approved):
LGTM, but there is a merge conflict now

PR #96, opened 2020-11-20
PMM-6863 Moved pxc/psmdb k8s/api and k8s/apimachinery to k8_api

askomorokhov:
https://jira.percona.com/browse/PMM-6863

https://github.com/percona-platform/dbaas-controller/pull/96
https://github.com/Percona-Lab/pmm-submodules/pull/1260

During the conversation with @AlekSi and @BupycHuk decided to use `string` instead of `k8s.io/apimachinery/pkg/api/resource/quantity.Quantity`
https://github.com/kubernetes/apimachinery/blob/671277bf9349218735d59b748a5c0b41076f9469/pkg/api/resource/quantity.go#L89-L100


AlekSi (changes requested):
Please remove all +k8s, +optional, etc comments – we don't have tooling that uses them

AlekSi (changes requested):
Please also remove protobuf tags. And see comment about inline JSON in the ticket.

BupycHuk (approved):

BupycHuk:
can we exclude k8api directory from linter?

askomorokhov:
> can we exclude k8api directory from linter?

Yes, we can. It was excluded. But then I have added it back - as we going to "maintain" this code.

AlekSi (changes requested):
We should keep only types we directly use with json tags, without inline tag, without methods. I believe I said that before.

We should not use Quantity (that's just a string for us).
We don't use RawExtension.

k8s_api/apimachinery/pkg/apis/meta/v1/types.go
```diff
@@ -0,0 +1,336 @@
...
+// ObjectMeta is metadata that all persisted resources must have, which includes all objects
+// users must create.
+type ObjectMeta struct {
+	// Name must be unique within a namespace. Is required when creating resources, although
+	// some resources may allow a client to request the generation of an appropriate name
+	// automatically. Name is primarily intended for creation idempotence and configuration
+	// definition.
+	// Cannot be updated.
+	// More info: http://kubernetes.io/docs/user-guide/identifiers#names
+	Name string `json:"name,omitempty"`
```
AlekSi, line 95: I believe that's the only field we use. Let's remove all others

askomorokhov:
@AlekSi, 

> We should not use Quantity (that's just a string for us).
This is the breaking change for UI - expects numbers. Should we break it in this PR or as a separate tasks UI/BE? 
How we should validate input? Format on UI?

AlekSi:
dbaas-controller doesn't have UI. pmm-managed is responsible for making dbaas-controller API calls with valid parameters

BupycHuk:
@AlekSi we will have to move this quantity parsing logic to pmm-managed or to UI anyway. From my point of view it looks like out of scope of this task because it changes behaviour of API.

k8s_api/apimachinery/pkg/apis/meta/v1/types.go
askomorokhov: and labels.

.golangci-required.yml
```diff
@@ -10,6 +10,14 @@ linters-settings:
       - errors
       # use "github.com/golang/protobuf/proto" instead
       - github.com/gogo/protobuf/proto
+      # use github.com/percona-platform/dbaas-controller/k8s_api/api/core/v1 instead https://jira.percona.com/browse/PMM-6863
+      - k8s.io/api/core/v1
+      # use k8s.io/apimachinery/pkg/... instead https://jira.percona.com/browse/PMM-6863
```
AlekSi, line 15: That comment is wrong now

cmd/dbaas-controller/main.go
```diff
@@ -26,7 +26,8 @@ import (
 	"google.golang.org/grpc/grpclog"
 	"gopkg.in/alecthomas/kingpin.v2"
 
-	_ "github.com/percona-platform/dbaas-controller/catalog" // load messages
+	// load messages.
+	_ "github.com/percona-platform/dbaas-controller/catalog"
```
AlekSi, line 30: No, that comment should be on that line – it explains the blank import

k8s_api/apimachinery/pkg/apis/meta/v1/types.go
```diff
@@ -0,0 +1,94 @@
...
+
+	// Map of string keys and values that can be used to organize and categorize
+	// (scope and select) objects. May match selectors of replication controllers
+	// and services.
+	// More info: http://kubernetes.io/docs/user-guide/labels
+	Labels map[string]string `json:"labels,omitempty" protobuf:"bytes,11,rep,name=labels"`
```
AlekSi, line 93: Please drop protobuf tags too

k8s_api/api/core/v1/types.go
```diff
@@ -0,0 +1,72 @@
...
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+// Package v1 contains type from k8s.io/api/core/v1.
+package v1
+
+import (
+	metav1 "github.com/percona-platform/dbaas-controller/k8s_api/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
```
BupycHuk, line 24: do we use these constants?

k8s_api/api/core/v1/types.go
```diff
@@ -0,0 +1,72 @@
...
+	NamespaceAll string = ""
+	// NamespaceNodeLease is the namespace where we place node lease objects (used for node heartbeats).
+	NamespaceNodeLease string = "kube-node-lease"
+	// TopologyKeyAny is the service topology key that matches any node.
+	TopologyKeyAny string = "*"
+)
+
+// HostPathVolumeSource represents a host path mapped into a pod.
+// Host path volumes do not support ownership management or SELinux relabeling.
+type HostPathVolumeSource struct {
```
BupycHuk, line 37: Do we use it?

k8s_api/pxc/v1/pxc_backup_types.go
```diff
@@ -17,19 +17,19 @@
...
-	metav1.TypeMeta `json:",inline"`
+	metav1.TypeMeta
 	metav1.ListMeta `json:"metadata"`
 	Items           []PerconaXtraDBClusterBackup `json:"items"`
 }
 
 // PerconaXtraDBClusterBackup represents a Percona XtraDB cluster backup.
 type PerconaXtraDBClusterBackup struct {
-	metav1.TypeMeta   `json:",inline"`
+	metav1.TypeMeta
```
AlekSi, line 32: It should have JSON tag, but not ",inline"

k8s_api/psmdb/v1/psmdb_types_test.go
```diff
@@ -0,0 +1,219 @@
...
+            "image": "percona/percona-server-mongodb-operator:1.4.0-backup",
+            "serviceAccountName": "percona-server-mongodb-operator"
+        },
+        "pmm": {}
+    },
+    "status": {}
+}
+`
+
+func TestPSMDBTypesMarshal(t *testing.T) {
```
BupycHuk, line 119: Can we add the same test for XtraDB?

AlekSi (changes requested):

k8s_api/pxc/v1/pxc_backup_types.go
askomorokhov: I'll take a look

k8s_api/api/core/v1/types.go
askomorokhov: Only here https://github.com/percona-platform/dbaas-controller/blob/main/k8_api/common/types.go#L34

@BupycHuk, should I remove both?

cmd/dbaas-controller/main.go
askomorokhov: Yeah, it was incorrectly formatted by `golangci-lint run --fix`

k8s_api/pxc/v1/pxc_backup_types.go
askomorokhov: @AlekSi, can you please suggest what tag it should have? 

this is inline field and its fields have JSON tags.

```
type TypeMeta struct {
	// Kind is a string value representing the REST resource this object represents.
	// Servers may infer this from the endpoint the client submits requests to.
	// Cannot be updated.
	// In CamelCase.
	// More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds
	Kind string `json:"kind,omitempty"`

	// APIVersion defines the versioned schema of this representation of an object.
	// Servers should convert recognized schemas to the latest internal value, and
	// may reject unrecognized values.
	// More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources
	APIVersion string `json:"apiVersion,omitempty"`
}
```

AlekSi: > Anonymous struct fields are usually marshaled as if their inner exported fields were fields in the outer struct, subject to the usual Go visibility rules amended as described in the next paragraph. An anonymous struct field with a name given in its JSON tag is treated as having that name, rather than being anonymous.

You are absolutely right, it should not have a tag.
Given how much time we spend on this inline problem, let's add comment like `// anonymous for embedding` to not forget about it :)

k8s_api/api/core/v1/types.go
askomorokhov: Removed

.golangci-required.yml
askomorokhov: Changed.

cmd/dbaas-controller/main.go
askomorokhov: fixed.

k8s_api/apimachinery/pkg/apis/meta/v1/types.go
askomorokhov: Dropped.

k8s_api/api/core/v1/types.go
BupycHuk: let's keep

service/k8sclient/k8client.go
```diff
@@ -351,19 +349,15 @@ func (c *K8Client) getPerconaXtraDBClusters(ctx context.Context) ([]XtraDBCluste
 
 // getDeletingClusters returns clusters which are not fully deleted yet.
 func (c *K8Client) getDeletingClusters(ctx context.Context, managedBy string, runningClusters map[string]struct{}) ([]Cluster, error) {
-	var list meta.List
+	var list psmdb.PerconaServerMongoDBDeletingList
```
BupycHuk, line 352: it can be both type of pods

askomorokhov: Could you please explain what mean?
In k8s only one type of pod https://github.com/kubernetes/api/blob/master/core/v1/types.go#L3662-L3683

service/cluster/xtra_db_cluster.go
```diff
@@ -68,22 +69,22 @@ func (s *XtraDBClusterService) ListXtraDBClusters(ctx context.Context, req *cont
 		params := &controllerv1beta1.XtraDBClusterParams{
 			ClusterSize: cluster.Size,
 			Pxc: &controllerv1beta1.XtraDBClusterParams_PXC{
-				DiskSize: cluster.PXC.DiskSize,
+				DiskSize: convertors.StrToBytes(cluster.PXC.DiskSize),
```
BupycHuk, line 72: isn't it better to move calling converters inside k8client?

BupycHuk: To not leave a chance to use wrong suffixes

askomorokhov: I've considered such case - but decided to call it here as we convert values in communication between dbaas-controller and pmm-managed but not with Kubernetes API.

AlekSi:
Move k8s_api to `service/k8sclient/internal` to ensure that those package are not leaked outside of k8sclient. PTAL

AlekSi (approved):

service/k8sclient/k8client.go
BupycHuk: list type is `psmdb.PerconaServerMongoDBDeletingList`, so it's better to use common type.

BupycHuk: `getDeletingClusters` can be called for both cluster types

askomorokhov: Renamed PerconaServerMongoDBDeletingList to PodList

BupycHuk (approved):

BupycHuk (approved):

PR #101, opened 2020-12-03
PMM-6854 Return status changing for PXC if status is empty

percona-csalguero:
When the XtraDB cluster is starting, Kubernetes returns an empty status that's replicated on the front end.
This tiket changes that behavior and returns status `Changing` if the status is empty.

|Component|Link|
|-----|-----|
|Jira|https://jira.percona.com/browse/PMM-6854|
|FB|https://github.com/Percona-Lab/pmm-submodules/pull/1285|


askomorokhov (approved):
Looks good to me.

BupycHuk (approved):

service/k8sclient/k8client.go
```diff
@@ -349,6 +349,14 @@ func (c *K8Client) getPerconaXtraDBClusters(ctx context.Context) ([]XtraDBCluste
 	return res, nil
 }
 
+func getPXCState(state pxc.AppState) ClusterState {
+	clusterState, ok := pxcStatesMap[state]
+	if !ok {
```
BupycHuk, line 354: The problem with this approach that status in operators can be unknown.

AlekSi:
@percona-csalguero Please resolve merge conflicts

AlekSi (commented):
@percona-csalguero @BupycHuk As an idea: do we have an information about CR creation time? If yes, can we return Changing state for empty / unknown / unexpected status only if CR is less that, say, 10 seconds old?

I'm afraid about about the fact we hide unknown / unprotected statuses without even logging.

BupycHuk (approved):

PR #135, opened 2021-01-27
PMM-7250 Return NotFound on Get when cluster is deleted, FailedPrecondition when it's not ready, Rename Get -> GetCredentials

jprukner:
https://jira.percona.com/browse/PMM-7250

Right now we call Get on every cluster. What it actually returns is credentials to connect to the db cluster. That is why we decided to rename this API endpoint to GetCredentials.
As part of changes on frontend side we should call it only on demand, thus avoiding errors when user does not have sufficient permissions to view credentials.
We want to avoid returning credentials on changing cluster so this PR covers returning FailedPrecondition in this case.
We also want to return NotFound when cluster is deleting or just non-existent.


packages.dot
```diff
@@ -1,5 +1,6 @@
 digraph packages {
 	"/service/cluster" -> "/service/k8sclient";
+	"/service/cluster" -> "/service/k8sclient/kubectl";
```
AlekSi, line 3: May we avoid that dependency on the implementation details?

AlekSi: See https://github.com/percona-platform/dbaas-controller/pull/137

jprukner: :+1: I removed the dependency.

service/k8sclient/k8sclient_test.go
```diff
@@ -49,6 +49,16 @@ func TestK8sClient(t *testing.T) {
 
 	l := logger.Get(ctx)
 
+	t.Run("Get non-existing clusters", func(t *testing.T) {
+		t.Parallel()
+		if _, err := client.GetPSMDBCluster(ctx, "d0ca1166b638c-psmdb"); !errors.Is(err, ErrNotFound) {
```
BupycHuk, line 54: let's use `assert.EqualError`

BupycHuk: here and in other places

service/k8sclient/k8sclient.go
```diff
@@ -522,13 +522,23 @@ func (c *K8sClient) DeleteXtraDBCluster(ctx context.Context, name string) error
 	return nil
 }
 
+// ErrNotFound should be returned when referenced resource does not exist
+// inside Kubernetes cluster.
+var ErrNotFound error = errors.New("resource was not found in Kubernetes cluster")
```
BupycHuk, line 527: let's move this to other errors on the top

service/k8sclient/k8sclient.go
```diff
@@ -522,13 +522,23 @@ func (c *K8sClient) DeleteXtraDBCluster(ctx context.Context, name string) error
...
+
 // GetXtraDBCluster returns an XtraDB cluster credentials.
 func (c *K8sClient) GetXtraDBCluster(ctx context.Context, name string) (*XtraDBCredentials, error) {
 	var cluster pxc.PerconaXtraDBCluster
 	err := c.kubeCtl.Get(ctx, string(perconaXtraDBClusterKind), name, &cluster)
 	if err != nil {
+		if errors.Is(err, kubectl.ErrNotFound) {
+			return nil, errors.Wrap(ErrNotFound, "cannot get XtraDb cluster")
```
BupycHuk, line 535: `cannot get XtraDB cluster credentials`

BupycHuk (changes requested):

service/k8sclient/k8sclient_test.go
jprukner: done

service/k8sclient/k8sclient.go
jprukner: done

service/k8sclient/k8sclient.go
jprukner: I created const error template, used both for xtradb and mongo.

AlekSi (changes requested):
CI fails hard

jprukner:
> CI fails hard

I downgraded minikube version. Tests are working now.

service/k8sclient/k8sclient.go
```diff
@@ -551,7 +565,6 @@ func (c *K8sClient) GetXtraDBCluster(ctx context.Context, name string) (*XtraDBC
 	return credentials, nil
 }
 
-// GetXtraDBCluster returns an XtraDB cluster credentials.
```
BupycHuk, line 554: 👍🏼

BupycHuk (approved):

askomorokhov (approved):
Looks good to me. 👍

.github/workflows/ci.yml
```diff
@@ -86,6 +86,9 @@ jobs:
       - name: Configure minikube
         run: |
           sudo apt install -y conntrack
+          sudo curl -Lo /tmp/minikube https://github.com/kubernetes/minikube/releases/download/v1.16.0/minikube-linux-amd64
+          sudo chmod +x /tmp/minikube
+          sudo mv /tmp/minikube $(which minikube)
```
AlekSi, line 91: Please switch to using https://github.com/percona-platform/actions-setup-minikube and adding minikube version to `strategy.matrix`

AlekSi (approved):
LGTM with a suggestion. Please do not re-request review from me for this PR :)

.github/workflows/ci.yml
AlekSi: (if that's easy to do in this PR)

jprukner: I will do this in separate PR.

artemgavrilov (approved):

jprukner:
rebased on pmm/2.15

PR #145, opened 2021-02-16
PMM-6884 Use EKS in CI for DBaaS Controller

jprukner:
Add EKS for tests.


AlekSi:
> Replace minkube

Not sure we should _replace_ it – we still use minikube locally. I think it is better to _add_ EKS (as a separate job?).

AlekSi:
Oh, and `KUBERNETES_VERSION` is not actually used after those changes

jprukner:
> > Replace minkube
> 
> Not sure we should _replace_ it – we still use minikube locally. I think it is better to _add_ EKS (as a separate job?).

I think the point of this ticket is to eliminate minikube setup phase to make tests run faster.

AlekSi:
The point is to get feedback faster – if tests fail on EKS, no need to wait for minikube tests to finish. Another point is to just test with EKS – it might be different from minikube in some aspects we care about. But we still need to test with minikube

jprukner:
> The point is to get feedback faster – if tests fail on EKS, no need to wait for minikube tests to finish. Another point is to just test with EKS – it might be different from minikube in some aspects we care about. But we still need to test with minikube. We also need to test with different k8s versions, and currently that's not possible with EKS.

That's fair. I will then add eks as another job.

.github/workflows/ci.yml
```diff
@@ -178,6 +178,166 @@ jobs:
...
+        with:
+          file: crosscover.out
+          flags: crosscover
+          env_vars: GO_VERSION,KUBERNETES_VERSION
+          fail_ci_if_error: false
+      
+      - name: Delete the test namespace
+        if: ${{ always() }}
+        run: |
+          kubectl config get-contexts $(kubectl config current-context)
```
jprukner, line 319: debug

Makefile
```diff
@@ -152,18 +151,30 @@ collect-debugdata:                ## Collect debugdata
 	minikube kubectl -- logs --all-containers --timestamps --selector='name=percona-server-mongodb-operator' > ./debugdata/psmdb-operators.txt
 	minikube kubectl -- logs --all-containers --timestamps --selector='app.kubernetes.io/name=percona-server-mongodb' > ./debugdata/psmdb-clusters.txt
 
+eks-setup-test-namespace:
+	kubectl ${KUBECTL_ARGS} create ns "${NAMESPACE}"
+	kubectl ${KUBECTL_ARGS} config set-context --current --namespace="${NAMESPACE}"
+	kubectl ${KUBECTL_ARGS} config get-contexts $(kubectl ${KUBECTL_ARGS} config current-context) | sed 's/.*\(dbaas-controller-test-[0-9a-z]*\).*/\1/' | tail -1
```
jprukner, line 157: debug

.github/workflows/ci.yml
```diff
@@ -178,6 +178,166 @@ jobs:
...
+
+      - name: Deploy operators to the test namespace
+        run: |
+          ls "${HOME}/.kube/config"
+          kubectl config get-contexts $(kubectl config current-context)
+          make eks-install-operators
+
+      - name: Run tests
+        run: |
+          kubectl --kubeconfig ${HOME}/.kube/config version -o json
```
jprukner, line 304: delete debug

.github/workflows/ci.yml
```diff
@@ -178,6 +178,166 @@ jobs:
...
+           export RANDOM_STRING=$(head -500 /dev/urandom | tr -dc 'a-z0-9' | fold -w 16 | tail -n +1 | head -n 1)
+           export NAMESPACE="dbaas-controller-test-${RANDOM_STRING}"
+           make eks-setup-test-namespace NAMESPACE="${NAMESPACE}"
+        env:
+           BASE64_KUBECONFIG: ${{secrets.BASE64_KUBECONFIG}}
+
+      - name: Deploy operators to the test namespace
+        run: |
+          ls "${HOME}/.kube/config"
+          kubectl config get-contexts $(kubectl config current-context)
```
jprukner, line 299: delete debug, this line and line above

.github/workflows/ci.yml
```diff
@@ -178,6 +178,162 @@ jobs:
...
+        with:
+          file: crosscover.out
+          flags: crosscover
+          env_vars: GO_VERSION,KUBERNETES_VERSION
+          fail_ci_if_error: false
+      
+      - name: Delete the test namespace
+        if: ${{ always() }}
+        run: |
+          make eks-delete-current-namespace
```
percona-csalguero, line 316: In my experience (not much) if we don't delete the operators there are garbage left.
For example, network interfaces, cloud formations, etc because those resources are owned by the operator so only deleting the namespace won't be enough.
Can we test that and add a line to uninstall operators first?

percona-csalguero (changes requested):

.github/workflows/ci.yml
jprukner: Let me investigate this. I thought when one deletes namespace all its resources die with it.

.github/workflows/ci.yml
```diff
@@ -178,6 +178,162 @@ jobs:
...
+
+      - name: Check that API tests can be built
+        run: make test-api-build
+
+      - name: Setup environment for tests to run
+        run: |
+           export PATH_TO_KUBECONFIG="${HOME}/.kube/config"
+           mkdir -p "${PATH_TO_KUBECONFIG%/*}"
+           echo "${BASE64_KUBECONFIG}" > "${PATH_TO_KUBECONFIG}"
+           export RANDOM_STRING=$(head -500 /dev/urandom | tr -dc 'a-z0-9' | fold -w 16 | tail -n +1 | head -n 1)
```
askomorokhov, line 290: May work as well: `export RANDOM_STRING=$(cat /proc/sys/kernel/random/uuid)`

.github/workflows/ci.yml
```diff
@@ -178,6 +178,162 @@ jobs:
...
+        run: |
+          # Break job if any files were changed during its run (code generation, etc), except go.sum.
+          # `go mod tidy` could remove old checksums from that file, and that's okay on CI,
+          # and actually expected for PRs made by @dependabot.
+          # Checksums of actually used modules are checked by previous `go` subcommands.
+          pushd tools && go mod tidy && git checkout go.sum
+          popd        && go mod tidy && git checkout go.sum
+          git diff --exit-code
+
+      - name: Run debug commands on failure
```
askomorokhov, line 328: 👍

askomorokhov (approved):
Overall looks good to me 👍 

But I have the same warnings as Carlos has. Deleting the namespace may not delete all artefacts.

What do you think about to display at the end of the execution some resource statistics related to the Kubernetes cluster: count of namespaces, volumes, volumes claims, etc? So we can see trends and react earlier f something?

.github/workflows/ci.yml
jprukner: Did not find any issues with network interfaces not being deleted but proper clean up won't do any harm. I will delete operators.

Makefile
```diff
@@ -152,18 +151,28 @@ collect-debugdata:                ## Collect debugdata
...
+	cat ./deploy/pxc-secrets.yaml | sed "s/pmmserver:.*/pmmserver: ${PMM_PASS}/g" | kubectl ${KUBECTL_ARGS} delete -f -
 	# Delete the PSMDB operator
-	cat ./deploy/psmdb-operator.yaml | ${KUBECTL_CMD} delete -f -
-	cat ./deploy/psmdb-secrets.yaml | sed "s/PMM_SERVER_USER:.*$/PMM_SERVER_USER: ${PMM_USER}/g;s/PMM_SERVER_PASSWORD:.*=$/PMM_SERVER_PASSWORD: ${PMM_PASS}/g;" | ${KUBECTL_CMD} delete -f -
+	cat ./deploy/psmdb-operator.yaml | kubectl ${KUBECTL_ARGS} delete -f -
+	cat ./deploy/psmdb-secrets.yaml | sed "s/PMM_SERVER_USER:.*/PMM_SERVER_USER: ${PMM_USER}/g;s/PMM_SERVER_PASSWORD:.*/PMM_SERVER_PASSWORD: ${PMM_PASS}/g;" | kubectl ${KUBECTL_ARGS} delete -f -
+
+eks-delete-current-namespace:
+	NAMESPACE=$$(kubectl ${KUBECTL_ARGS} config get-contexts $(kubectl ${KUBECTL_ARGS} config current-context) | sed 's/.*\(dbaas-controller-test-[0-9a-z]*\).*/\1/' | tail -1); \
+	if [ "$$NAMESPACE" != "default" ]; then kubectl delete ns "$$NAMESPACE"; fi
```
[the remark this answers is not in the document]
jprukner: :+1: Good suggestion. I will use it, thanks.

jprukner:
> Overall looks good to me +1
> 
> But I have the same warnings as Carlos has. Deleting the namespace may not delete all artefacts.
> 
> What do you think about to display at the end of the execution some resource statistics related to the Kubernetes cluster: count of namespaces, volumes, volumes claims, etc? So we can see trends and react earlier f something?

@askomorokhov I added this step:
```
 - name: Report Kubernetes resources count and state
        if: steps.setupenv.outputs.exit_code == 0
        run: |
          echo "Kubernetes resources counts:"
          echo "Namespaces: $(kubectl get ns | tail +2 | wc -l)"
          echo "Persistent volumes: $(kubectl get pv | tail +2 | wc -l)"
          echo "Pods: $(kubectl get pod | tail +2 | wc -l)"
          echo "State of all pods:"
          kubectl describe pods
```

.github/workflows/ci.yml
jprukner: Deleting of operator deployments added, also secrets are being deleted now. Can't delete CRD because they are cluster scoped, it would break other tests running in parallel.

percona-csalguero (approved):

PR #159, opened 2021-03-17
PMM-7701 Remove special chars from passwords

jprukner:
PSMDB operator does not handle well special characters in the password. We remove them for now until K8SPSMDB-364 is fixed.

https://github.com/Percona-Lab/pmm-submodules/pull/1616

### DIFF
--- a/service/k8sclient/k8sclient.go
+++ b/service/k8sclient/k8sclient.go
@@ -19,9 +19,10 @@ package k8sclient
 
 import (
 	"context"
+	"crypto/rand"
 	"encoding/json"
 	"fmt"
-	"math/rand"
+	"math/big"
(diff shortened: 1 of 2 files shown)


service/k8sclient/k8sclient.go
```diff
@@ -19,9 +19,10 @@ package k8sclient
 
 import (
 	"context"
+	"crypto/rand"
```
jprukner, line 22: I believe it's much more secure to use crypto/rand than math/rand.

BupycHuk (approved):

percona-csalguero (approved):

percona-csalguero (approved):

PR #171, opened 2021-04-06
PMM-7819 add HAProxy

jprukner:


service/k8sclient/k8sclient_test.go
```diff
@@ -82,7 +83,7 @@ func TestK8sClient(t *testing.T) {
 
 		l.Info("No XtraDB Clusters running")
 
-		err = client.CreateXtraDBCluster(ctx, &XtraDBParams{
+		err := client.CreateXtraDBCluster(ctx, &XtraDBParams{
```
jprukner, line 86: We need this to avoid data race on err variable. We run psmdb and pxc tests in parallel to speed up tests.

service/k8sclient/k8sclient.go
```diff
@@ -664,13 +697,14 @@ func (c *K8sClient) RestartXtraDBCluster(ctx context.Context, name string) error
 		return err
 	}
 
-	// TODO: implement logic to handle the case when there is HAProxy instead of ProxySQL.
-	_, err = c.kubeCtl.Run(ctx, c.restartDBClusterCmd(name, "proxysql"), nil)
-	if err != nil {
-		return err
+	for _, proxy := range []string{"proxysql", "haproxy"} {
+		if _, err := c.kubeCtl.Run(ctx, []string{"get", "statefulset", name + "-" + proxy}, nil); err == nil {
```
BupycHuk, line 701: will this command fail if there is no proxysql or haproxy?

service/k8sclient/k8sclient.go
```diff
@@ -685,17 +719,13 @@ func (c *K8sClient) getPerconaXtraDBClusters(ctx context.Context) ([]XtraDBClust
 	for i, cluster := range list.Items {
 		val := XtraDBCluster{
 			Name:    cluster.Name,
-			Size:    cluster.Spec.ProxySQL.Size,
+			Size:    cluster.Spec.PXC.Size,
```
BupycHuk, line 722: 👍🏼

BupycHuk (commented):
Looks good to me, only one question and one minor request.
Could you also update build tags to pass all tests, please?

service/cluster/xtra_db_cluster.go
```diff
@@ -147,14 +167,21 @@ func (s *XtraDBClusterService) CreateXtraDBCluster(ctx context.Context, req *con
...
+	}
+	if req.Params.Proxysql != nil {
+		params.ProxySQL = &k8sclient.ProxySQL{
 			Image:            req.Params.Proxysql.Image,
 			ComputeResources: computeResources(req.Params.Proxysql.ComputeResources),
 			DiskSize:         convertors.BytesToStr(req.Params.Proxysql.DiskSize),
-		},
-		PMMPublicAddress: req.PmmPublicAddress,
+		}
+	} else {
```
palash25, line 179: this else branch can be removed using this 

```go
if req.Params.Proxysql != nil {
	params.ProxySQL = &k8sclient.ProxySQL{
		Image:            req.Params.Proxysql.Image,
		ComputeResources: computeResources(req.Params.Proxysql.ComputeResources),
		DiskSize:         convertors.BytesToStr(req.Params.Proxysql.DiskSize),
	}

	err = client.CreateXtraDBCluster(ctx, params)
	if err != nil {
		return nil, status.Error(codes.Internal, err.Error())
    }
	return new(controllerv1beta1.CreateXtraDBClusterResponse), nil
}

params.HAProxy = &k8sclient.HAProxy{
	Image:            req.Params.Haproxy.Image,
	ComputeResources: computeResources(req.Params.Haproxy.ComputeResources),
}
err = client.CreateXtraDBCluster(ctx, params)
if err != nil {
	return nil, status.Error(codes.Internal, err.Error())
}
return new(controllerv1beta1.CreateXtraDBClusterResponse), nil
```

service/cluster/xtra_db_cluster.go
```diff
@@ -80,38 +76,62 @@ func (s *XtraDBClusterService) ListXtraDBClusters(ctx context.Context, req *cont
...
+				memoryBytes, err := convertors.StrToBytes(cluster.ProxySQL.ComputeResources.MemoryBytes)
+				if err != nil {
+					return nil, status.Error(codes.Internal, err.Error())
+				}
+				params.Proxysql.ComputeResources = &controllerv1beta1.ComputeResources{
+					CpuM:        int32(cpuMillis),
+					MemoryBytes: int64(memoryBytes),
+				}
+			}
+		} else if cluster.HAProxy.ComputeResources != nil {
```
palash25, line 103: maybe we should not use `else if` if there is no `else` at the end, just  use two seperate if conditions

service/cluster/xtra_db_cluster.go
```diff
@@ -80,38 +76,62 @@ func (s *XtraDBClusterService) ListXtraDBClusters(ctx context.Context, req *cont
...
 			},
-			Proxysql: &controllerv1beta1.XtraDBClusterParams_ProxySQL{
-				DiskSize: int64(proxySQLDiskSize),
-			},
 		}
-		if cluster.PXC.ComputeResources != nil {
-			cpuMillis, err := convertors.StrToMilliCPU(cluster.PXC.ComputeResources.CPUM)
+
+		if cluster.ProxySQL != nil {
+			proxySQLDiskSize, err := convertors.StrToBytes(cluster.ProxySQL.DiskSize)
```
palash25, line 82: this code is quite hard to read, even if we can't solve the nested nature of this bit the least we can do to improve readability is move all the code under the if condition to a separate functions for e.g. `setProxySQLParams` or `setHAProxyParams`

service/k8sclient/k8sclient.go
```diff
@@ -704,6 +734,16 @@ func (c *K8sClient) getPerconaXtraDBClusters(ctx context.Context) ([]XtraDBClust
 				{size: cluster.Status.PXC.Size, ready: cluster.Status.PXC.Ready},
 			},
 		}
+		if cluster.Spec.ProxySQL != nil {
```
palash25, line 737: we can get rid of the else branch here too

```go

if cluster.Spec.ProxySQL != nil {
	val.ProxySQL = &ProxySQL{
		DiskSize:         c.getDiskSize(cluster.Spec.ProxySQL.VolumeSpec),
		ComputeResources: c.getComputeResources(cluster.Spec.ProxySQL.Resources),
	}
    res[i] = val
    return res, nil
}

val.HAProxy = &HAProxy{
	ComputeResources: c.getComputeResources(cluster.Spec.HAProxy.Resources),
}
res[i] = val
return res, nil
````

File: service/k8sclient/k8sclient.go (hunk 2 above)
Comment (jprukner) [reply service/k8sclient/k8sclient.go:701]: You mean no proxy at all? Then the one gets `return errors.New("failed to restart pxc cluster proxy statefulset")`.

When there is only HAProxy it fails with not found for ProxySQL and vice versa.

Comment (BupycHuk) [reply service/k8sclient/k8sclient.go:701]: ok, good

File: service/cluster/xtra_db_cluster.go (hunk 4 above)
Comment (BupycHuk) [reply service/cluster/xtra_db_cluster.go:179]: this code will be harder to read and goes against makes us duplicate code.

File: service/k8sclient/k8sclient.go (hunk 7 above)
Comment (jprukner) [reply service/k8sclient/k8sclient.go:737]: I used `continue` to get rid of the else branch.

File: service/cluster/xtra_db_cluster.go (hunk 5 above)
Comment (jprukner) [reply service/cluster/xtra_db_cluster.go]: :+1: Done.

File: service/cluster/xtra_db_cluster.go (hunk 6 above)
Comment (jprukner) [reply service/cluster/xtra_db_cluster.go]: Done

File: service/cluster/xtra_db_cluster.go (hunk 4 above)
Comment (jprukner) [reply service/cluster/xtra_db_cluster.go:206]: @palash25 If it was one line of code we duplicate and `else` block was larger I'd do it. But like this, I think it's not worth it.

File: service/cluster/xtra_db_cluster.go [hunk 8]
```diff
@@ -50,6 +50,58 @@ func NewXtraDBClusterService(p *message.Printer) *XtraDBClusterService {
...
+		return err
+	}
+	memoryBytes, err := convertors.StrToBytes(proxy.ComputeResources.MemoryBytes)
+	if err != nil {
+		return err
+	}
+	params.Proxysql.ComputeResources = &controllerv1beta1.ComputeResources{
+		CpuM:        int32(cpuMillis),
+		MemoryBytes: int64(memoryBytes),
+	}
```
BupycHuk, line 78: looks like it can be extracted and reused

jprukner: Done

service/cluster/xtra_db_cluster.go
jprukner: I ended up refactoring them out. But the `if` blocks are now shorter.

palash25 (approved):

BupycHuk (approved):

service/k8sclient/k8sclient.go
```diff
@@ -543,6 +572,10 @@ func (c *K8sClient) UpdateXtraDBCluster(ctx context.Context, params *XtraDBParam
 		cluster.Spec.ProxySQL.Resources = c.updateComputeResources(params.ProxySQL.ComputeResources, cluster.Spec.ProxySQL.Resources)
 	}
 
+	if params.HAProxy != nil {
+		cluster.Spec.HAProxy.Resources = c.updateComputeResources(params.HAProxy.ComputeResources, cluster.Spec.HAProxy.Resources)
```
BupycHuk, line 576: probably we should check if `cluster.Spec.HAProxy` is exists and return error if it's not. 
Also we should check if cluster is not trying to run proxySQL and HAProxy in the same time.

palash25 (approved):

PR #173, opened 2021-04-12
PMM-7353 Pass PMM Credentials via API.

BupycHuk:
https://github.com/Percona-Lab/pmm-submodules/pull/1677
https://github.com/percona/pmm-managed/pull/723


service/k8sclient/k8sclient.go
```diff
@@ -402,26 +412,22 @@ func (c *K8sClient) CreateXtraDBCluster(ctx context.Context, params *XtraDBParam
...
 	}
 
 	// TODO: add a link to ticket to set random password for all other users.
-	data := secret.Data
-	data["root"] = []byte(pwd)
-
-	err = c.CreateSecret(ctx, secretName, data)
-	if err != nil {
-		return errors.Wrap(err, "cannot create secret for PXC")
+	secrets := map[string][]byte{
```
palash25, line 422: can you explain this a bit in comments?
i can see that its mapping the same password to different names, are these different components of an XtraDBCluster that share the same password?

BupycHuk: That's correct and there is comment above to add a link to ticket to set random password for all other users.

palash25: also should we call the variable `rootPassword` if it is being mapped to things other than `root`? maybe just call it pwd?

palash25: > can you explain this a bit in comments?
i can see that its mapping the same password to different names, are these different components of an XtraDBCluster that share the same password?


okay I get it it represents the contents of this file that was deleted `pxc-secrets.yaml ` so maybe a comment saying `// secrets is equivalent to the contents of a pxc-secrets.yaml file used to create a cluster ` something like that

BupycHuk: yes, we still should call it `rootPassword` to make it clear that we use root password for all secrets. and it should be fixed

service/k8sclient/k8sclient.go
```diff
@@ -801,24 +818,21 @@ func (c *K8sClient) CreatePSMDBCluster(ctx context.Context, params *PSMDBParams)
...
 		return errors.Wrap(err, "failed to generate password")
 	}
 
-	data := secret.Data
-	data["MONGODB_USER_ADMIN_PASSWORD"] = []byte(pwd)
-
-	err = c.CreateSecret(ctx, secretName, data)
-	if err != nil {
-		return errors.Wrap(err, "cannot create secret for PXC")
+	secrets := map[string][]byte{
```
palash25, line 827: same as https://github.com/percona-platform/dbaas-controller/pull/173/files#r612711109

service/k8sclient/k8sclient.go
BupycHuk: > okay I get it it represents the contents of this file that was deleted pxc-secrets.yaml so maybe a comment saying // secrets is equivalent to the contents of a pxc-secrets.yaml file used to create a cluster something like that

Will do it.

palash25 (approved):

jprukner (approved):

service/k8sclient/k8sclient.go
```diff
@@ -167,15 +167,25 @@ type Replicaset struct {
 	DiskSize         string
 }
 
+// PMM contains information related to PMM.
```
oter, line 170: Maybe better to specify what kind of information it contains?

oter (approved):

jprukner (approved):

service/k8sclient/k8sclient.go
```diff
@@ -82,17 +82,17 @@ const (
...
+	psmdbBackupImage    = "percona/percona-server-mongodb-operator:1.7.0-backup"
+	psmdbDefaultImage   = "percona/percona-server-mongodb:4.2.8-8"
+	psmdbAPIVersion     = "psmdb.percona.com/v1-7-0"
+	psmdbSecretNameTmpl = "dbaas-%s-psmdb-secrets"
 
 	// Max size of volume for AWS Elastic Block Storage service is 16TiB.
-	maxVolumeSizeEBS uint64 = 16 * 1024 * 1024 * 1024 * 1024
+	maxVolumeSizeEBS   uint64 = 16 * 1024 * 1024 * 1024 * 1024
+	internalSecretTmpl        = "internal-%s"
+	pullPolicy                = common.PullIfNotPresent
```
jprukner: :+1:

jprukner (approved):

palash25 (approved):

PR #175, opened 2021-04-22
PMM-7839 don't use saas build to test the controller against

jprukner:
dbaas-controller is not included in saas, don't run tests with that build tag.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,14 +26,12 @@ jobs:
           # https://docs.aws.amazon.com/eks/latest/userguide/kubernetes-versions.html
           - 1.15.11
           - 1.16.8
-        build_tags: ["", "saas"]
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -26,14 +26,12 @@ jobs:
           # https://docs.aws.amazon.com/eks/latest/userguide/kubernetes-versions.html
           - 1.15.11
           - 1.16.8
-        build_tags: ["", "saas"]
```
BupycHuk, line 29: I would just remove "saas" from here and that would be the best way to include it back if required.

BupycHuk (approved):

.github/workflows/ci.yml
jprukner: ok

PR #179, opened 2021-05-04
PMM-3019 add deployment for minikube

denisok:
To easy deploy one minikube cluster instead of separate PMM Server in
docker container.
```
$ minikube start --cpus 12 --memory 32G --driver=kvm2
$ minikube kubectl -- apply -f pmm-server-minikube.yaml
$ minikube service pmm-deployment
```

minikube will open browser with 2 opened ports.

This is mostly for testing purposes.

Add `pmm` as IP address in PMM. BTW @BupycHuk is there a way to pass IP as env variable ?


jprukner (approved):
It took some ~6mins to init the pod but it's working well.

denisok:
> It took some ~6mins to init the pod but it's working well.

yep, we need to separate `/srv` preparation from install, update and etc, so it will take much faster

jprukner:
the new env will start working as soon as the pmm-server:2 points to 2.17, when it's released.

denisok:
> the new env will start working as soon as the pmm-server:2 points to 2.17, when it's released.

OK, should be May 10th. so far I use old one locally, and it make sense to add new one right away.

BupycHuk (approved):

PR #188, opened 2021-05-20
PMM-7737 Generate password for each user.

BupycHuk:
https://github.com/Percona-Lab/pmm-submodules/pull/1793

### DIFF
--- a/service/k8sclient/k8sclient.go
+++ b/service/k8sclient/k8sclient.go
@@ -19,10 +19,8 @@ package k8sclient
 
 import (
 	"context"
-	"crypto/rand"
 	"encoding/json"
 	"fmt"
-	"math/big"
 	"net/http"
 	"reflect"
 	"strings"
@@ -48,7 +46,6 @@ type ClusterKind string
 const (
 	perconaXtraDBClusterKind = ClusterKind("PerconaXtraDBCluster")
 	perconaServerMongoDBKind = ClusterKind("PerconaServerMongoDB")
-	passwordLength           = 24
 )
 
 // ClusterState represents XtraDB cluster CR state.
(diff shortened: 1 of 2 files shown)


service/k8sclient/k8sclient.go
```diff
@@ -560,6 +548,45 @@ func (c *K8sClient) CreateXtraDBCluster(ctx context.Context, params *XtraDBParam
 	return c.kubeCtl.Apply(ctx, res)
 }
 
+func (c *K8sClient) generateXtraDBPasswords() (map[string][]byte, error) {
+	root, err := generatePassword(passwordLength)
```
oter, line 552: I would suggest to generate 6 passwords using a loop and then just use that array for assigning.

service/k8sclient/k8sclient.go
```diff
@@ -1069,6 +1082,39 @@ func (c *K8sClient) CreatePSMDBCluster(ctx context.Context, params *PSMDBParams)
...
+	}
+	userAdmin, err := generatePassword(passwordLength)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to generate userAdmin password")
+	}
+
+	// secrets represents stringData part of
+	// https://github.com/percona/percona-xtradb-cluster-operator/blob/main/deploy/secrets.yaml.
+	secrets := map[string][]byte{
+		"MONGODB_BACKUP_USER":              []byte("backup"),
```
jprukner, line 1106: similar here, I'd set only the password keys in the beginning, doing the same as for generateXtraDBPasswords, with small change -> adding *_USER keys befor returning. We may even reuse some code with couple of more changes. 
Something like
```

func generatePasswords(map[string][]byte) error {...}
```

jprukner (changes requested):
Let's DRY this code :)

jprukner (approved):

service/k8sclient/secrets.go
```diff
@@ -0,0 +1,93 @@
...
+func generatePSMDBPasswords() (map[string][]byte, error) {
+	// secrets represents stringData part of
+	// https://github.com/percona/percona-xtradb-cluster-operator/blob/main/deploy/secrets.yaml.
+	secrets := map[string][]byte{
+		"MONGODB_BACKUP_USER":          []byte("backup"),
+		"MONGODB_CLUSTER_ADMIN_USER":   []byte("clusterAdmin"),
+		"MONGODB_CLUSTER_MONITOR_USER": []byte("clusterMonitor"),
+		"MONGODB_USER_ADMIN_USER":      []byte("userAdmin"),
+	}
+	passwords, err := generatePasswords(map[string][]byte{
```
oter, line 80: it's a bit confusing to me having `[]byte` and not using it. Let's replace it with `map[string]struct{}`.

BupycHuk: I reuse the same map to generate a new one.

oter (approved):

askomorokhov (approved):

PR #193, opened 2021-06-10
PMM-8174 install operator rpc now handles given version

jprukner:
Install opreator calls now have version and fetch manifests from github instead of using embedded manifest files.
We can pass url from which manifests are fetched -> air-gaped clusters can run their fake github user content site.
EDIT:
I had to split install and update feature because applying bundle.yaml - we do that for operator installation - may not work for all upgrades. Instead I apply new CRDs and RBAC and patch the operator deployment to use correct image.


service/k8sclient/k8sclient.go
```diff
@@ -1319,45 +1345,40 @@ func (c *K8sClient) CheckOperators(ctx context.Context) (*Operators, error) {
...
 
 // checkOperatorStatus returns if operator is installed and operators version.
 // It checks for all API versions supported by the operator and based on the latest API version in the list
 // figures out which version of operator is installed.
-func (c *K8sClient) checkOperatorStatus(installedVersions []string, expectedAPIVersion string) (operator Operator) {
-	apiNamespace := strings.Split(expectedAPIVersion, "/")[0]
+func (c *K8sClient) checkOperatorStatus(installedVersions []string, apiPrefix string) (operator Operator) {
 	operator.Status = OperatorStatusNotInstalled
 	lastVersion, _ := version.NewVersion("v0.0.0")
+	zeroVersion := lastVersion
```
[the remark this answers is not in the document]
jprukner: nope

service/k8sclient/k8sclient.go
```diff
@@ -72,21 +72,21 @@ const (
...
-	psmdbCRVersion      = "1.8.0"
-	psmdbBackupImage    = "percona/percona-server-mongodb-operator:1.8.0-backup"
-	psmdbDefaultImage   = "percona/percona-server-mongodb:4.2.8-8"
-	psmdbAPIVersion     = "psmdb.percona.com/v1-8-0"
-	psmdbSecretNameTmpl = "dbaas-%s-psmdb-secrets"
+	psmdbBackupImageTemplate = "percona/percona-server-mongodb-operator:%s-backup"
+	psmdbDefaultImage        = "percona/percona-server-mongodb:4.2.8-8"
+	psmdbAPINamespace        = "psmdb.percona.com"
+	psmdbAPIVersionTemplate  = psmdbAPINamespace + "/v%s"
+	psmdbSecretNameTmpl      = "dbaas-%s-psmdb-secrets"
```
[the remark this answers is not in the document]
jprukner: nope

service/k8sclient/internal/kubectl/kubectl.go
```diff
@@ -244,7 +244,9 @@ func (k *KubeCtl) Run(ctx context.Context, args []string, stdin interface{}) ([]
 func run(ctx context.Context, kubectlCmd []string, args []string, stdin interface{}) ([]byte, error) {
 	l := logger.Get(ctx)
 	l = l.WithField("component", "kubectl")
-	args = append(kubectlCmd, args...)
+	cmds := make([]string, len(kubectlCmd))
+	copy(cmds, kubectlCmd)
+	args = append(cmds, args...)
```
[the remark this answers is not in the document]
jprukner: What's wrong with that?

service/k8sclient/k8sclient.go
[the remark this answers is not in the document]
jprukner: considered and nope

service/k8sclient/internal/kubectl/kubectl.go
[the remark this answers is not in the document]
BupycHuk: why did you change it?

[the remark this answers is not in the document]
jprukner: Because it was source of data race on k.cmd. When you ran two parallel kubectl commands on the same kubectl object, it races.

service/k8sclient/k8sclient.go
```diff
@@ -1319,45 +1342,40 @@ func (c *K8sClient) CheckOperators(ctx context.Context) (*Operators, error) {
...
+		if newVersion.GreaterThan(lastVersion) {
+			lastVersion = newVersion
 		}
-		lastVersion = newVersion
-		operator.Version = lastVersion.String()
+	}
+	operator.Version = lastVersion.String()
+	if lastVersion != zeroVersion { // comparing pointers
+		operator.Status = OperatorStatusOK
 	}
```
BupycHuk, line 1379: we should check if operator version exist in version service and return ok if it is, or unsupported if it's not.

BupycHuk (changes requested):

service/k8sclient/k8sclient.go
```diff
@@ -118,15 +116,14 @@ type OperatorStatus int32
 const (
 	// OperatorStatusOK represents that operators are installed and have supported API version.
 	OperatorStatusOK OperatorStatus = 1
-	// OperatorStatusUnsupported represents that operators are installed, but doesn't have supported API version.
-	OperatorStatusUnsupported OperatorStatus = 2
```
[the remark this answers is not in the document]
jprukner: It is but we can't decide inside dbaas-controller, I'd have to call version service also inside dbaas-controller. We return if it's installed and in which version. Then inside pmm-managed we decide if it's supported or not.

service/k8sclient/internal/kubectl/kubectl.go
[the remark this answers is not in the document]
jprukner: https://play.golang.org/p/pCWrXr5urKS
Run it with -race option.
`go run -race ...`
When cap is sufficient the modification happens on the slice you are appending into.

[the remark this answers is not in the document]
jprukner: So I copy the slice before appending into it to make sure the race does not happen.

service/k8sclient/k8sclient.go
jprukner: Is it good idea to request version service on two places like that?

service/k8sclient/internal/kubectl/kubectl.go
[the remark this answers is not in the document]
BupycHuk: it's so weird, but ok.

service/k8sclient/k8sclient.go
```diff
@@ -854,6 +867,10 @@ func (c *K8sClient) ListPSMDBClusters(ctx context.Context) ([]PSMDBCluster, erro
 
 // CreatePSMDBCluster creates percona server for mongodb cluster with provided parameters.
 func (c *K8sClient) CreatePSMDBCluster(ctx context.Context, params *PSMDBParams) error {
+	if params.Image == "" {
+		return errors.New("PSMDB image can't be empyt")
```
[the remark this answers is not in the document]
jprukner: Ok, that is fair. I will change it back and keep the old default images.

service/k8sclient/k8sclient.go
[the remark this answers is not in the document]
BupycHuk: Yeah, we should have whole logic of version comparison in one place instead of two.

service/k8sclient/k8sclient.go
BupycHuk: Then we should move it from here to pmm-managed or do opposite.

jprukner: I mean having two places where you have to maintain the abstractions.

service/k8sclient/k8sclient.go
```diff
@@ -584,16 +597,20 @@ func (c *K8sClient) UpdateXtraDBCluster(ctx context.Context, params *XtraDBParam
...
 // DeleteXtraDBCluster deletes Percona XtraDB cluster with provided name.
 func (c *K8sClient) DeleteXtraDBCluster(ctx context.Context, name string) error {
+	operators, err := c.CheckOperators(ctx)
+	if err != nil {
+		return err
+	}
 	res := &pxc.PerconaXtraDBCluster{
 		TypeMeta: common.TypeMeta{
-			APIVersion: pxcAPIVersion,
+			APIVersion: c.getAPIVersionForPXCOperator(operators),
```
jprukner, line 606: THIS HAS TO BE VERSION OF THE OPERATOR WE DEPLOYED THE CLUSTER WITH, RIGHT? 
TODO: CHECK THIS.

service/k8sclient/k8sclient.go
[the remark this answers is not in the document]
jprukner: todo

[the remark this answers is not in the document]
jprukner: done

service/k8sclient/k8sclient.go
```diff
@@ -1662,45 +1677,68 @@ func (c *K8sClient) doAPIRequest(ctx context.Context, method, endpoint string, o
...
+	return fmt.Sprintf(psmdbAPIVersionTemplate, strings.ReplaceAll(operators.Psmdb.Version, ".", "-"))
 }
 
-func (c *K8sClient) InstallXtraDBOperator(ctx context.Context) error {
-	file, err := dbaascontroller.DeployDir.ReadFile("deploy/pxc-operator.yaml")
-	if err != nil {
-		return err
-	}
-	return c.kubeCtl.Apply(ctx, file)
+func (c *K8sClient) getAPIVersionForPXCOperator(operators *Operators) string {
```
jprukner, line 1706: does it make sense to have this func, if I use it only at one place?

palash25: i don't really mind having a 1liner method because it makes the code easier to read at the call site rather than reading a long print statement

dbaas_controller.go
```diff
@@ -16,10 +16,3 @@
 
 // Package dbaascontroller is a dummy package.
 package dbaascontroller
-
-import (
-	"embed"
-)
-
-//go:embed deploy/*
```
palash25, line 24: why do we need this file now, if we have removed the build directive?

dbaas_controller.go
```diff
@@ -16,10 +16,3 @@
 
 // Package dbaascontroller is a dummy package.
 package dbaascontroller
-
```
BupycHuk, line 19: we will have conflict because of these changes, let's keep it.

service/cluster/kubernetes_cluster.go
```diff
@@ -31,7 +31,6 @@ import (
 
 var operatorStatusesMap = map[k8sclient.OperatorStatus]controllerv1beta1.OperatorsStatus{
```
BupycHuk, line 32: This status is always used with version, so if we don't have `Unsupported` status, then let's remove status at all and use empty `version` as `NotInstalled`

utils/app/app.go
```diff
@@ -58,6 +60,7 @@ func Setup(opts *SetupOpts) (*Flags, error) {
 	var flags Flags
 	kingpin.Flag("grpc.addr", "gRPC listen address").Default(":20201").StringVar(&flags.GRPCAddr)
 	kingpin.Flag("debug.addr", "Debug listen address").Default(":20203").StringVar(&flags.DebugAddr)
+	kingpin.Flag("github.url", "github url to fetch operator manifests from").Default("https://raw.githubusercontent.com").StringVar(&flags.GithubUserContentURL)
```
BupycHuk, line 63: why do we make this flag? if to make it easy to configure path to operator crds, why don't include `/percona/`? or even pass `url template`?

dbaas_controller.go
jprukner: I brought it back, to avoid conflicts as Nurlan suggested.
But anyway there si dbaas_controller_test.go, go refuses to run tests for packages that don't have non-test code. That is why there was this dummy file.

dbaas_controller.go
jprukner: done

service/cluster/kubernetes_cluster.go
jprukner: done

utils/app/app.go
jprukner: The template is a good idea, more flexible. I will implement it.

jprukner: Flag for each operator introduced.

service/k8sclient/k8sclient.go
```diff
@@ -586,7 +590,7 @@ func (c *K8sClient) UpdateXtraDBCluster(ctx context.Context, params *XtraDBParam
 func (c *K8sClient) DeleteXtraDBCluster(ctx context.Context, name string) error {
 	res := &pxc.PerconaXtraDBCluster{
 		TypeMeta: common.TypeMeta{
-			APIVersion: pxcAPIVersion,
+			APIVersion: pxcAPINamespace + "/v1",
```
BupycHuk, line 593: could you please check this case:
1. Add DB Cluster
2. Upgrade Operator version
3. Delete DB Cluster

service/k8sclient/k8sclient.go
```diff
@@ -112,33 +112,23 @@ const (
...
-	// OperatorStatusUnsupported represents that operators are installed, but doesn't have supported API version.
-	OperatorStatusUnsupported OperatorStatus = 2
-	// OperatorStatusNotInstalled represents that operators are not installed.
-	OperatorStatusNotInstalled OperatorStatus = 3
-)
-
 const (
 	clusterWithSameNameExistsErrTemplate = "Cluster '%s' already exists"
 	canNotGetCredentialsErrTemplate      = "cannot get %s cluster credentials"
+	operatorManifestsURLTemplate         = "%s/percona/%s/v%s/deploy/bundle.yaml"
```
BupycHuk, line 118: it's not used

BupycHuk (approved):

BupycHuk:
@jprukner tests are failing

palash25 (approved):

service/k8sclient/k8sclient.go
jprukner: deleted

service/k8sclient/k8sclient.go
jprukner: I use it only for deleting because it does not matter. It can't be empty and non-existing API. In upgrade docs it's mentioned it should be changed in all db clusters when upgrading operator so we should keep it in `Add DB Cluster` and patch DB clusters on operator upgrade -> #207

BupycHuk (approved):

PR #198, opened 2021-06-30
PMM-7805 Use Kubernetes API directly instead of running kubectl proxy

jprukner:
Use kuberenetes API directly instead of launching kubectl proxy.


BupycHuk (commented):
Do we still need proxy being run?

service/k8sclient/k8sclient.go
```diff
@@ -1622,13 +1623,25 @@ func (c *K8sClient) GetConsumedDiskBytes(ctx context.Context, clusterType kubern
 		if err != nil {
 			return 0, errors.Wrap(err, "can't compute consumed disk size: failed to get worker nodes")
 		}
+		// TODO We want to use clientcmd.Load in the future to load the kubeconfig from slice of bytes.
```
BupycHuk, line 1626: we do have kubeconfig at the moment we create k8sclient, so let's reuse it instead of reading from file.

jprukner: > Do we still need proxy being run?

We don't run it, do we? I removed all the code related to it, I think.

BupycHuk: yeah, you are right, missed that you removed proxy.go

jprukner: > we do have kubeconfig at the moment we create k8sclient, so let's reuse it instead of reading from file.

Thanks for reminding me, I got confused.

service/k8sclient/k8sclient.go
```diff
@@ -1622,13 +1623,27 @@ func (c *K8sClient) GetConsumedDiskBytes(ctx context.Context, clusterType kubern
...
+			return 0, errors.Wrap(err, "failed to build kubeconfig out of given path")
+		}
+		config, err := clientConfig.ClientConfig()
+		if err != nil {
+			return 0, errors.Wrap(err, "failed to build kubeconfig out of given path")
+		}
+		clientset, err := kubernetes.NewForConfig(config)
+		if err != nil {
+			return 0, errors.Wrap(err, "failed to build client out of submited kubeconfig")
+		}
```
BupycHuk, line 1637: What if we move it to `New`?

BupycHuk (approved):
Looks good to me, just one suggestion.

oter (approved):

service/k8sclient/k8sclient.go
jprukner: I would do it if we move to client-go based approach entirely. Right now it'd be unnecessary overhead when we use it only in one place.

PR #202, opened 2021-07-14
PMM-8355 Set TTL Monitor to 60s to make cluster come up

jprukner:
TTL monitor 60s is set in official example [cr.yaml](https://github.com/percona/percona-server-mongodb-operator/blob/v1.9.0/deploy/cr.yaml), it was cause of our cluster not coming up.


tests/go.mod
```diff
@@ -9,11 +9,11 @@ go 1.16
...
 require (
-	github.com/percona-platform/dbaas-api v0.0.0-20210420115836-fa7ca800ec9b
+	github.com/percona-platform/dbaas-api v0.0.0-20210630090346-f95d2ec8b7c1
 	github.com/pkg/errors v0.9.1
 	github.com/sirupsen/logrus v1.8.0
 	github.com/stretchr/testify v1.7.0
 	golang.org/x/net v0.0.0-20200520004742-59133d7f0dd7 // indirect
 	golang.org/x/sys v0.0.0-20201029080932-201ba4db2418
-	google.golang.org/grpc v1.36.1
+	google.golang.org/grpc v1.38.0
```
oter, line 18: is it needed for the task to update grpc version?

jprukner: it is not, but result of `go mod tidy`, I would need to do it anyway. Is there some issue with the version?

jprukner: Whatever, I reverted it, updated only dbaas-api which was blocking the tests.

oter: Technically update of other deps should be in a separate PR. You could use `go get github.com/percona-platform/dbaas-api@branch-name` for updating just needed dependency.

oter: fixed

jprukner: It turned out it is needed. Because dbaas-api uses it. I suppose all `go mod tidy` additions are needed for this to build.

oter (approved):

BupycHuk (approved):

PR #203, opened 2021-07-14
PMM-8408 Use correct pmm-client image tag

jprukner:
Use correct pmm-client image tag based on current PMM version.


denisok:
ok, fine for prod. But we still need to use `dev-latest` in development, how we would do that?

jprukner:
> ok, fine for prod. But we still need to use `dev-latest` in development, how we would do that?

good point, I added branching on the startup which chooses the right image

service/k8sclient/k8sclient.go
```diff
@@ -326,12 +325,35 @@ var (
...
 
+var pmmClientImage string
+
 // K8sClient is a client for Kubernetes.
 type K8sClient struct {
 	kubeCtl *kubectl.KubeCtl
 	l       logger.Logger
 }
 
+func init() {
```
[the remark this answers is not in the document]
jprukner: Would it be better to move this to `k8sclients.New`?
Advantage is we need to decide only once when placed in init. It does not matter I think because version does not change on runtime.

service/k8sclient/k8sclient.go
```diff
@@ -326,12 +325,35 @@ var (
...
+	}
+
+	v, err := goversion.NewVersion(pmmversion.PMMVersion)
+	if err != nil {
+		panic("failed to decide what version of pmm-client to use: " + err.Error())
+	}
+
+	if v.Core().String() == v.String() {
+		// Production version contains only major.minor.patch ...
+		pmmClientImage = "perconalab/pmm-client:2"
```
denisok, line 350: https://hub.docker.com/r/percona/pmm-client

should be `percona/pmm-client:2`

denisok (changes requested):

service/k8sclient/k8sclient.go
jprukner: done

BupycHuk (approved):

askomorokhov (approved):

service/k8sclient/k8sclient.go
denisok: don't see new commits so far

jprukner: Sorry, I did not noticed it failed to push. Now it's done done.

denisok (approved):

PR #204, opened 2021-07-14
PMM-8420 Add variable to disable deleting cluster

jprukner:
To not to leave garbage in EKS cluster, we should delete even on failure. One can turn off deleting of the test namespace by setting environment variable `DELETE_NAMESPACE` to `false`.


tests/go.mod
```diff
@@ -9,11 +9,11 @@ go 1.16
...
 require (
-	github.com/percona-platform/dbaas-api v0.0.0-20210420115836-fa7ca800ec9b
+	github.com/percona-platform/dbaas-api v0.0.0-20210630090346-f95d2ec8b7c1
 	github.com/pkg/errors v0.9.1
 	github.com/sirupsen/logrus v1.8.0
 	github.com/stretchr/testify v1.7.0
 	golang.org/x/net v0.0.0-20200520004742-59133d7f0dd7 // indirect
 	golang.org/x/sys v0.0.0-20201029080932-201ba4db2418
-	google.golang.org/grpc v1.36.1
+	google.golang.org/grpc v1.38.0
```
jprukner, line 18: just to make api test build

Dasio (approved):
Just I don't understand `always()` if it returns always true, what the point of it?

jprukner:
> Just I don't understand `always()` if it returns always true, what the point of it?

It has to be there, otherwise the job won't even evaluate the if statement.

askomorokhov (approved):

BupycHuk (approved):

PR #207, opened 2021-07-19
PMM-8462 Patch each db cluster after operator update

jprukner:
We need to change the versions in database CRs after updating the operator because if we don't the cluster is operated as if it is older CRVersion -> this would prevent updating database to the latest version.

https://www.percona.com/doc/kubernetes-operator-for-pxc/update.html

TODO: CHANGE THE BASE TO MAIN AFTER PMM-8462 IS MERGED

EDIT:
I needed to use real patch instead of apply, it was not working for every case. Hence the changes.


service/operator/xtradb_operator.go
```diff
@@ -44,10 +44,22 @@ func (x XtraDBOperatorService) InstallXtraDBOperator(ctx context.Context, req *c
...
 	if err != nil {
 		return nil, err
 	}
 
+	err = client.ApplyOperator(ctx, req.Version, x.manifestsURLTemplate)
+	if err != nil {
+		return nil, err
+	}
+
+	if operators.XtradbOperatorVersion != "" {
```
jprukner, line 57: Should we rely on this?
There could be a situation when no operator is installed but there are clusters deployed. But it's very corner case situation. Most likely the cluster won't be working properly long term anyway.

palash25: i am not sure about that but I think it would be a good idea to document that concern/edge case in a comment (here and in the psmdb code) at the very least

palash25 (approved):
LGTM, please fix the CI

oter (approved):

service/operator/xtradb_operator.go
jprukner: Note added.

jprukner:
rebased

service/k8sclient/internal/pxc/pxc_types.go
```diff
@@ -143,7 +143,7 @@ type PerconaXtraDBClusterList struct {
 type PodSpec struct { //nolint:maligned
 	Enabled                       bool                            `json:"enabled,omitempty"`
 	Pause                         bool                            `json:"pause,omitempty"`
-	Size                          int32                           `json:"size,omitempty"`
+	Size                          *int32                          `json:"size,omitempty"`
```
BupycHuk, line 146: I though that `omitempty` ignores field if it has default value, doesn't it?

percona-csalguero: IIRC, omitempty means that it won't be included in the json by the marshaler but if you want to **detect** if it was present or not, you need a pointer

BupycHuk: We never want to decrease size to 0. We use `pause` field for that, so we don't have to differentiate 0 and nil.

service/k8sclient/internal/psmdb/psmdb_types.go
```diff
@@ -29,8 +29,8 @@ type PerconaServerMongoDB struct {
 	common.TypeMeta   // anonymous for embedding
 	common.ObjectMeta `json:"metadata,omitempty"`
 
-	Spec   PerconaServerMongoDBSpec   `json:"spec,omitempty"`
-	Status perconaServerMongoDBStatus `json:"status,omitempty"`
+	Spec   *PerconaServerMongoDBSpec   `json:"spec,omitempty"`
+	Status *perconaServerMongoDBStatus `json:"status,omitempty"`
```
percona-csalguero, line 33: This won't encode/decode the json field since it is not exported.

percona-csalguero (changes requested):

oter (approved):

percona-csalguero (approved):

BupycHuk (approved):

PR #209, opened 2021-07-22
PMM-7914 Upgrade database

jprukner:
I replaced applying with patching in Update RPCs because it's more reliable. Apply can fail when CR gets updated concurentlly.

I use semiautomatic upgrade of databases that is setting update strategy to Rolling and changing the image of the cluster.

The new status Upgrading is returned if CR version does not match version in all of the pods.


service/k8sclient/k8sclient.go
```diff
@@ -1615,21 +1701,19 @@ func getResources(resources common.ResourceList) (cpuMillis uint64, memoryBytes
 	return cpuMillis, memoryBytes, nil
 }
 
-// GetConsumedCPUAndMemory returns consumed CPU and Memory in given namespace. If namespaces
-// is empty slice, it tries to get them from all namespaces.
-func (c *K8sClient) GetConsumedCPUAndMemory(ctx context.Context, namespaces ...string) (
+// GetConsumedCPUAndMemory returns consumed CPU and Memory in given namespace. If namespace
+// is empty, it tries to get them from all namespaces.
+func (c *K8sClient) GetConsumedCPUAndMemory(ctx context.Context, namespace string) (
```
jprukner, line 1706: I stumbled on this error while fixing GetPods method. There is no need for multiple namespaces. Even if we passed more namespaces to the old implementation, it would not work because kubectl supports working only with one namespace at a time.

JiriCtvrtka (approved):

palash25 (approved):

PR #222, opened 2021-09-17
PMM-8836 Use latest operators from version service instead of hardcoded ones

jprukner:
I forgot to change tests when upgrade of operators feature was implemented.

I copy&pasted code from pmm-managed as it was the easiest way to bring version service into dbaas-controller.
It does not feel like a good solutions though.
@BupycHuk @oter would you recommend importing code from pmm-managed? I does not feel like good solution either.

### DIFF
--- a/service/k8sclient/k8sclient_test.go
+++ b/service/k8sclient/k8sclient_test.go
@@ -24,6 +24,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"net/http"
 	"os"
 	"strings"
 	"testing"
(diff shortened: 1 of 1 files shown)


oter (commented):
IMO it would be better to have a library in a separate repository containing such clients and common code. But, looking at your case, you are copy-pasted the client for testing purposes, so probably we can keep your approach.

service/k8sclient/k8sclient_test.go
```diff
@@ -39,6 +41,140 @@ import (
...
+}
+
+type matrix struct {
+	PXCOperator   map[string]interface{} `json:"pxcOperator,omitempty"`
+	PSMDBOperator map[string]interface{} `json:"psmdbOperator,omitempty"`
+}
+
+type Version struct {
+	Product        string `json:"product"`
+	ProductVersion string `json:"operator"`
```
oter, line 73: Is that correct? The field is `ProductVersion`, but annotation `operator`

service/k8sclient/k8sclient_test.go
```diff
@@ -39,6 +41,140 @@ import (
...
+	return latest, nil
+}
+
+// Matrix calls version service with given params and returns components matrix.
+func (c *VersionServiceClient) Matrix(ctx context.Context, params componentsParams) (*VersionServiceResponse, error) {
+	paths := []string{c.url, params.product}
+	if params.productVersion != "" {
+		paths = append(paths, params.productVersion)
+	}
+	url := strings.Join(paths, "/")
```
oter, line 124: I would suggest to try `path.Join`

service/k8sclient/k8sclient_test.go
```diff
@@ -39,6 +41,140 @@ import (
...
+	req, err := http.NewRequestWithContext(ctx, "GET", url, nil)
```
oter, line 125: Let's use http.MethodGet

service/k8sclient/k8sclient_test.go
```diff
@@ -39,6 +41,140 @@ import (
...
+		return nil, err
+	}
+	resp, err := c.http.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	defer func() {
+		err := resp.Body.Close()
+		if err != nil {
+			log.Fatal(err)
```
oter, line 136: not sure it is a good idea to call Fatal when closing the body, better to add this error to retuned one

service/k8sclient/k8sclient_test.go
```diff
@@ -39,6 +41,140 @@ import (
...
+	if len(resp.Versions) != 1 {
+		return nil, nil, nil // no deps for the PMM version passed to c.Matrix
+	}
+	pmmVersionDeps := resp.Versions[0]
+	latestPSMDBOperator, err := latest(pmmVersionDeps.Matrix.PSMDBOperator)
+	if err != nil {
+		return nil, nil, err
+	}
+	latestPXCOperator, err := latest(pmmVersionDeps.Matrix.PXCOperator)
+	return latestPXCOperator, latestPSMDBOperator, err
```
oter, line 175: the better option is to `return nil, nil, err` if err != nil, otherwise `return latestPXCOperator, latestPSMDBOperator, nil`

service/k8sclient/k8sclient_test.go
```diff
@@ -39,6 +41,140 @@ import (
...
+	ProductVersion string `json:"operator"`
+	Matrix         matrix `json:"matrix"`
+}
+
+// VersionServiceResponse represents response from version service API.
+type VersionServiceResponse struct {
+	Versions []Version `json:"versions"`
+}
+
+var errNoVersionsFound error = errors.New("no versions to compare current version with found")
```
oter, line 82: looks like errNoVersionsFound is not needed, because you are not handling this case in the other places and this error is private, so cannot be used outside of current package.

oter (changes requested):

service/k8sclient/k8sclient_test.go
jprukner: > looks like errNoVersionsFound is not needed, because you are not handling this case in the other places and this error is private, so cannot be used outside of current package.

I return this error on two places. If I remove this I will return the errors.New("no versions to compare current version with found") there - I will repeat myself. I don't think that is better.

service/k8sclient/k8sclient_test.go
jprukner: It's deffered function. But I agree that just logging it is better.

service/k8sclient/k8sclient_test.go
jprukner: Yes, it is correct. The version service was designed only with operators in mind. Then we decided to include PMM <-> operator dependencies. I did not change it to be backwards compatible.

service/k8sclient/k8sclient_test.go
```diff
@@ -39,6 +42,143 @@ import (
...
+			return nil, err
+		}
+		if parsedVersion.GreaterThan(latest) {
+			latest = parsedVersion
+		}
+	}
+	return latest, nil
+}
+
+func latestProduct(s []Version) (*goversion.Version, error) {
```
oter, line 102: is that func used somewhere else except the test?

jprukner: It is used only in tests.

oter: everything is good 👍

oter (approved):

artemgavrilov (approved):

PR #244, opened 2021-12-28
tools: Remove vulnerable indirect dependency

palash25:
fix for https://github.com/percona-platform/dbaas-controller/security/dependabot/tools/go.sum/github.com%2Fdgrijalva%2Fjwt-go/open


gen1us2k:
Thank you for your contribution. dbaas-controller uses v3.2.3+incompatible now

PR #305, opened 2022-07-06
PMM-9597 Added missing suffixes

gen1us2k:
Added missed suffixes for kibibytes and kilobytes in lowercase

PMM-9597

Build: SUBMODULES-0

- [ ] Links to other linked pull requests (optional).


utils/convertors/convertors.go
```diff
@@ -54,6 +54,8 @@ func StrToBytes(memory string) (uint64, error) {
 		"m":  0.001,
```
gen1us2k, line 54: @BupycHuk What is it stands for? milibytes?

BupycHuk: 😄 looks like we were trying to reuse the same function for CPU and memory and then when we split them forgot to remove this line.

gen1us2k: I left 'm' unit and added a comment for the future.

BupycHuk (approved):

percona-csalguero (approved):

PR #334, opened 2022-08-23
PMM-10562 Support CR override on cluster creation

gen1us2k:
PMM-10562

Build: SUBMODULES-0

- [ ] Links to other linked pull requests (optional).


BupycHuk (approved):

service/k8sclient/k8sclient.go
```diff
@@ -506,6 +512,24 @@ func (c *K8sClient) CreatePXCCluster(ctx context.Context, params *PXCParams) err
 			},
 		},
 	}
+	bytes, err := ioutil.ReadFile(pxcCRFile)
+	// Ignore error if we failed to open cr file
+	if err == nil {
```
denisok, line 517: probably we need a little bit more logging here. To log that we actually have overrides, and maybe as debug - what overrides (if that doesn't contain some secrets).

percona-csalguero (approved):

service/k8sclient/k8sclient.go
```diff
@@ -2242,3 +2152,300 @@ func (c *K8sClient) makeReq112Plus(params *PSMDBParams, extra extraCRParams) *ps
...
 	return req
 }
+
+func (c *K8sClient) createPSMDBSpec(operator *goversion.Version, params *PSMDBParams, extra *extraCRParams) (*psmdb.PerconaServerMongoDB, error) {
+	spec := new(psmdb.PerconaServerMongoDB)
+	bytes, err := ioutil.ReadFile(psmdbCRFile)
+	if err == nil {
+		err = c.unmarshalTemplate(bytes, spec)
+		if err != nil {
+			return nil, err
```
denisok, line 2162: do we need hard exit here ? Maybe just message that template isn't good and continue with default?
maybe better to get Max feedback - what he will prefer to test those templates as well...

denisok (approved):

PR #436, opened 2023-02-24
Added docker image check

percona-csalguero:
PMM-10687 DBaaS: DB cluster creation fails with unreleased version of pmm-server

Since the docker dependency is going to be added anyway for the self-upgrade feature @michal-kralik is working on, I used that module to determine if the image exists because we don't have the `-dev` suffix anymore in the `pmmversion.PMMVersion`  variable

Build: https://github.com/Percona-Lab/pmm-submodules/pull/3086

### DIFF
--- a/service/k8sclient/k8sclient.go
+++ b/service/k8sclient/k8sclient.go
@@ -30,6 +30,8 @@ import (
 	"time"
 
 	"github.com/AlekSi/pointer"
+	dockerTypes "github.com/docker/docker/api/types"
+	"github.com/docker/docker/client"
 	goversion "github.com/hashicorp/go-version"
 	psmdbv1 "github.com/percona/percona-server-mongodb-operator/pkg/apis/psmdb/v1"
 	pxcv1 "github.com/percona/percona-xtradb-cluster-operator/pkg/apis/pxc/v1"
@@ -79,6 +81,9 @@ const (
 	appStateStopping string = "stopping"
 	appStateReady    string = "ready"
 	appStateError    string = "error"
+
+	// Dev-latest docker image.
+	devLatest = "perconalab/pmm-client:dev-latest"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


service/k8sclient/k8sclient.go
```diff
@@ -346,6 +351,34 @@ func init() {
...
+}
+
+func imageExists(ctx context.Context, image string) (bool, error) {
+	cli, err := client.NewClientWithOpts(client.FromEnv, client.WithAPIVersionNegotiation())
+	if err != nil {
+		panic(err)
+	}
+	defer cli.Close() //nolint
+
+	reader, err := cli.ImagePull(ctx, image, dockerTypes.ImagePullOptions{})
```
recharte, line 370: Is there a way of checking if the image exists without pulling it? Seems like an unnecessary action.

percona-csalguero: It doesn't really pull. It returns a reader interface which is closed in the line below this without actually downloading the image.

recharte: ahhh ok great

recharte (approved):

service/k8sclient/k8sclient.go
```diff
@@ -346,6 +351,34 @@ func init() {
 	}
 
 	pmmClientImage = "percona/pmm-client:" + v.Core().String()
+
+	exists, err := imageExists(context.Background(), pmmClientImage)
+
+	// if !exists or there was an error while checking if the image exists, use dev-latest as default.
+	if !exists || err != nil {
```
michal-kralik, line 358: This seems pretty dangerous.
If there's intermittent issues with connection to the Docker registry, we will default to `dev-latest`.
I suppose this will be the case in production as well, correct?

What about defaulting to `pmmClientImageEnv` instead?
Since the ticket is meant to resolve issues during testing, I think this is a safer approach.

michal-kralik (changes requested):

